PR #8, opened 2019-04-30
Added simple FFD generation function to geo_utils (and a regression test for it)

nbons:


joanibal:
Thank you for writing the tests!

PR #16, opened 2019-07-31
General function to make design variable demonstration movie. Added argument to ref axis for raySize parameter.

nbons:


joanibal:
This looks really neat! Do you have an example for it?

PR #18, opened 2019-10-14
Fixed demoDesignVars to work with localSection DVs

A-Gray-94:
@anilyil, @nbons not sure which one of you is best placed to check this

Previously, when looping through design variables, the demoDesignVars method would check whether the variable appeared in `self.DV_listLocal` or `self.DV_listSectionLocal` and then either way try to get the variable bounds from `self.DV_listLocal` meaning you'd get a `KeyError` if you used `SectionLocal` variables.

### DIFF
--- a/DVGeometry.py
+++ b/DVGeometry.py
@@ -2129,11 +2129,16 @@ class DVGeometry(object):
         # Loop through design variables
         count = 0
         for key in dvDict:
-            if key in self.DV_listLocal or key in self.DV_listSectionLocal:
+            if key in self.DV_listLocal:
                 if not includeLocal:
                     continue
                 lower = self.DV_listLocal[key].lower
                 upper = self.DV_listLocal[key].upper
+            elif key in self.DV_listSectionLocal:
+                if not includeLocal:
+                    continue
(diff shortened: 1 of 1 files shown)


anilyil:
@nbons wrote that routine initially I believe. My modifications should only be related to indentation, but I might have messed up the routine since I don't know it that well. I did not understand your problem. Did you fix it? Is there a feature that is broken that used to work previously?

A-Gray-94:
Sorry, should have been clearer, this issue was nothing to do with your changes, I just thought I'd tag the two people who were working on pygeo the most at the moment. I guess since @nbons wrote the original function he is the best person to double check this fix.

nbons:
This looks fine. I made this change on my local repo among other things, just haven't got around to making a pull request.

A-Gray-94:
Do you have anything against merging this then? Or would you rather merge your fork instead?

nbons:
No, you can go ahead and merge it. Or Josh can. I don't have the power to
do so. I'll submit an additional pull request at some point.

A-Gray-94:
@joanibal

PR #23, opened 2019-12-09
Monotonic constraint, demoDesignVars, LE radius constraint

nbons:
- Added a parameter to the monotonic constraint so that you can select a subset of a design variable vector to which the monontonic constraint is applied.
- Fixed some problems with demoDesignVars. Most egregious bug was that before it was adding the dvs as deltas rather than just setting them directly, which didn't work for dvs with nonzero starting values. Also added default behavior to do a sinusoidal sweep going from initial to upper bound to lower bound and back to initial so that movies look smooth.
- Finally, added a constraint to handle leading edge radius requirements. The setup is the same as a 1D thickness constraint. You specify points along LE, the code calculates three points to define the circular arc and then computes radius. Derivatives done with complex step. Also added visualization so you can see the circles defining the LE radius.


bbrelje:
I removed myself as a reviewer because I am not familiar with these constraints and have never used them. I appreciate you fixing these bugs

Note that there is currently no Travis CI automated test coverage of DVConstraints (only DVGeo) so the passed test shouldn't suggest anything about whether this commit is breaking or not

joanibal:
@anilyil

DVGeometry.py
```diff
@@ -2102,11 +2113,14 @@ def demoDesignVars(self, directory, includeLocal=True, pointSet=None, callBack=N
         pointSet : str
             Name of the pointset to write out. If this is not specified, it will
             take the first one in the list.
-        callBack : function
+	    callBack : function
```
anilyil: can you convert tab to spaces here for consistency

DVGeometry.py
```diff
@@ -2124,58 +2138,66 @@ def demoDesignVars(self, directory, includeLocal=True, pointSet=None, callBack=N
                 raise Error('DVGeo must have a point set to update for\
                             demoDesignVars to work.')
         else:
-            writePointSet = True
+	        writePointSet = True
```
anilyil: can you convert tab to spaces here for consistency

DVGeometry.py
```diff
@@ -2124,58 +2138,66 @@ def demoDesignVars(self, directory, includeLocal=True, pointSet=None, callBack=N
...
+                    lower = geo.DV_listSectionLocal[key].lower
+                    upper = geo.DV_listSectionLocal[key].upper
+                elif key in geo.DV_listGlobal:
+                    if not includeGlobal:
+                        continue
+                    lower = geo.DV_listGlobal[key].lower
+                    upper = geo.DV_listGlobal[key].upper
+
+                x = dvDict[key].flatten()
+                nDV = len(lower)
```
anilyil: would this work if this DV did not have a lower bound?

DVConstraints.py
```diff
@@ -718,6 +717,157 @@ def addThicknessConstraints1D(self, ptList, nCon, axis,
...
+                                axis[0], axis[1], axis[2]))
+            coords[i, 0] = up
+            coords[i, 1] = down
+
+        # Calculate mid-points
+        midPts = (coords[:,0,:] + coords[:,1,:]) / 2.0
+
+        # Project to get leading edge point
+        lePts = numpy.zeros((nSpan, 3))
+        streamwise = numpy.array([-1.0, 0, 0])
```
anilyil: You are assuming stream-wise direction is in the -x direction. Is this a fair assumption? Some meshes might not have this property. Maybe you can include an optional parameter to pick this. It might also be useful when we are computing this constraint for swept wings for example. We might want to get the third point so that we are projecting orthogonally to the LE line, instead of just projecting in the x direction.

DVConstraints.py
```diff
@@ -2553,6 +2713,196 @@ def writeTecplot(self, handle):
...
+        xi_of_eta = numpy.einsum('ij,i->ij', nxi, eta_on_xi)
+
+        # Remove component of eta in the xi direction
+        neta = neta - xi_of_eta
+        for i in range(self.nCon):
+            neta[i] /= geo_utils.euclideanNorm(neta[i])
+
+        return origin, nxi, neta
+
+    def computeCircle(self, coords):
```
anilyil: Can you include some short text that describes the mathematical formulation here? A reference also works.

DVConstraints.py
```diff
@@ -718,6 +717,162 @@ def addThicknessConstraints1D(self, ptList, nCon, axis,
...
+                                axis[0], axis[1], axis[2]))
+            coords[i, 0] = up
+            coords[i, 1] = down
+
+        # Calculate mid-points
+        midPts = (coords[:,0,:] + coords[:,1,:]) / 2.0
+
+        # Project to get leading edge point
+        lePts = numpy.zeros((nSpan, 3))
+        chordDir /= numpy.linalg.norm(numpy.array(chordDir, 'd'))
```
anilyil: this wont work if chordDir is given as a list. I think you can also set the default to -1 0 0 as an array, I just wanted to say that having the option to modify this is good imo. You can keep where you are setting the value explicitly in the test you wrote

PR #36, opened 2020-06-12
Setup Install

joanibal:
## Purpose
Change the structure of the repo to use the `setup.py` method of installation. 
This will make it easier to install environment specific packages and specify dependencies. 

One can install the repo now by doing
`make` (if applicable) and `pip install .`

## Type of change
- Code style update (formatting, renaming)
- Refactoring (no functional changes, no API changes)

## Checklist


joanibal:
The tests fail for the current docker because dependencies required for install are not met because the other repos aren't updated.

Issue #40, opened 2020-09-07
Broken  `writePlot3dCoef()` with numpy >=1.18


marcomangano:
## Type of issue

- Bug

## Description
The function `writePlot3dCoef()` in `pyBlock.py` is broken when `numpy==1.18` is used. 
It looks like the function `numpy.flatten()` now only requires a string input to define the flattening order, and not an integer as we are currently doing (see lines 660, 662, 664 of `pyBlock.py`).

I believe the issue is generated by [this deprecated module](https://numpy.org/doc/1.18/release/1.18.0-notes.html#the-axis-argument-to-numpy-ma-mask-cols-and-numpy-ma-mask-row-is-deprecated)

### Steps to reproduce issue
1. `pip install numpy==1.18`
2. Run a script that uses `DVGeometry.writePlot3d()`

### Current behavior
Broken function, the run crashes throwing the following value error:
```
Traceback (most recent call last):
  File "Analysis.py", line 239, in <module>
    DVGeoc1.writePlot3d(outputDirectory + "/ffd_deformed_P%.0f.xyz" % (args.pitch))
  File "/home/mmangano/miniconda3/envs/pygeo-dev/lib/python3.8/site-packages/pygeo/DVGeometry.py", line 2113, in writePlot3d
    self.FFD.writePlot3dCoef(fileName)
  File "/home/mmangano/miniconda3/envs/pygeo-dev/lib/python3.8/site-packages/pygeo/pyBlock.py", line 660, in writePlot3dCoef
    vals[:, :, :, 0].flatten(1).tofile(f, sep="\n")
ValueError: Non-string object detected for the array ordering. Please pass in 'C', 'F', 'A', or 'K' instead
```
### Expected behavior
The function should replicate the behavior tested with `numpy` 1.16 and output a coordinate file with the correct ordering.

### Code version 
**pygeo:** 1.2.0

**Python version:** 3.8

**External dependencies:** numpy 1.18.5

PR #41, opened 2020-09-07
Fix `writePlot3dCoef()`

marcomangano:
## Purpose
Fixes the bug discussed in #40 

## Type of change

- Bugfix (non-breaking change which fixes an issue)

## Testing
Follow the steps described in the issue, now the code does not error out and the output file is ordered correctly to be read by Tecplot.

## Checklist

- [x] I have run unit and regression tests which pass locally with my changes (although that small function is probably not covered)


joanibal (approved):

marcomangano:
@bbrelje @joanibal this PR should be ready to go, it's a small but pretty useful fix!

bbrelje:
@marcomangano do you happen to know when the character flag was introduced? will this break backward compatibility with any versions of numpy that you know of?

marcomangano:
@bbrelje I briefly mention it in the linked issue. I think the character flag has been around for a few versions, but the code broke with `numpy>=1.18`. I tried the fix on `numpy==1.16` and it works. As that is the minimum requirement for our supported versions I do not foresee any issue, but it is your call in case you want to make a new release. 
We should probably add a test for this, but I believe there are other relevant parts of the code with higher priority, right?

bbrelje (approved):

nwu63 (approved):

PR #47, opened 2020-10-24
New style tests closes #44

### LINKED ISSUE #44: Update tests
# Description
- [ ] Update tests to use testflo
- [ ] Add coverage report since this is pure Python

bbrelje:
## Purpose
This PR updates all unit tests to the new testflo standard and adds a little bit of new coverage for DVconstraints. There is also a small bugfix that gets the blocks tests to run (they didn't since scipy 1.3.0)

## Type of change
What types of change is it?

- [ X] New feature (non-breaking change which adds functionality)
- [ ] Documentation update
- [ X] Maintenance update
- [ ] Other (please describe)

## Testing
All the old test data was ported over into the new file format so all the DVGeometry tests are producing the exact same results as before. Test coverage is significantly expanded and more maintainable going forward. 

## Checklist

- [X ] I have run unit and regression tests which pass locally with my changes
- [X] I have added new tests that prove my fix is effective or that my feature works


bbrelje:
This one is good to go @anilyil @joanibal

bbrelje:
I added some new tests and unfortunately it looks like the test 9 (colinearity constraint) derivatives are producing different values on different architectures. It passes on my machine. 

Our choices are:
- deprecate (is anyone using it?) and remove the test
- add a warning (use at your own risk) and remove the test
- someone else fixes the bug and we comment out the test for now
- someone else fixes the bug and we leave the test in, but ignore test 9 failures when accepting PRs

I don't have time to fix the underlying issue.

joanibal:
I vote for option 1. I don't see much value in maintaining features that aren't ever used. 

Could you post something in slack to see if anyone can think of a reason it should not be deprecated?

bbrelje:
Unfortunately the problem seems to be compiling pyspline with intel vs gcc. on my own machine with the latest pyspline, compiling with intel passes all the tests; compiling gcc fails tests 9 and 13b

We probably can't ignore it?

setup.py
```diff
@@ -60,7 +60,8 @@
       install_requires=[
             'numpy>=1.16',
             'pyspline>=1.1',
-            'scipy>=1.2'
+            'scipy>=1.2',
+            'numpy-stl'
```
anilyil, line 64: I was going to say that I needed to manually install this package to run the tests but I see you added them as a dependency. I did not need to re-run setup after pulling from your branch so not sure how we can fix this problem for other users who will just do a git pull and not do the setup again

.travis.yml
```diff
@@ -15,7 +15,7 @@ script:
   - |
     if [[ "$TEST_TYPE" == "docker" ]]; then
     # run tests
-      docker exec -it app /bin/bash -c ". $BASHRC && cd $DOCKER_WORKING_DIR/reg_tests && python run_reg_tests.py -nodiff"
+      docker exec -it app /bin/bash -c ". $BASHRC && cd $DOCKER_WORKING_DIR/tests && testflo . -v -n 1"
```
anilyil, line 18: I think we can run testflo with 2 procs, even though the tests themselves are set to run on a single proc. However, this might break some tests if they are trying to read/write the same files because this way, testflo would run 2 tests at a time, using a single processor for each test. If the tests do not write files that are used by other tests concurrently, it should be fine and we can use 2 procs, but @nwu63 would know more as the expert

anilyil:
Other than those two points, the tests look good and they run fine on my installation. Do you want me to check anything else in the tests themselves?

.travis.yml
bbrelje: The only test that writes a file that I know of is Nick's cylinder test which writes an xyz file. No other test writes that file. There are some simultaneous reads. I think it's probably fine. Will this break reg tests for users who don't have mpi4py installed and configured correctly?

bbrelje:
No, I don't have anything else in particular to review

setup.py
bbrelje: This only breaks tests if they don't have numpy-stl, not the package itself so I think it's OK

.travis.yml
eirikurj: In general for these test we should try to have unique filenames for each test. The file read should be fine. As @anilyil mentioned we should probably use `-n 2` here to try to accelerate testing on Travis since all test use `N_PROCS = 1` anyways. This is just for running with testflo on Travis so this does not break anything. 
If I recall correctly testflo will run even if mpi is not installed. The `--nompi` flag can also be supplied to not use mpi (for those who are configured to run in parallel). Also, if the user does not have testflo installed one can always just run the tests manually also.
Finally, please remember to update run command in the [runscript in the docker repo](https://github.com/mdolab/docker/blob/master/extras/master_test_only_updated_GFORTRAN.py) for the collective testing.

nwu63: I doubt it will speed up Travis, since what we actually get is 1 physical proc with 2 logical procs. I am in favour of keeping it `-n 1` here. On the build bot we can increase to `-n 4` since there are more procs available.

setup.py
nwu63: IMO we shouldn't get all users to install this package by default. The proper way is to have `optional_dependencies` for testing, for example
```
optional_dependencies = {
    'testing': [
        'numpy-stl',
    ],
}
```
and you will install all the required plus the `testing` optional dependencies if you use `pip install . [testing]`. (Can someone test that this actually works, this is what OpenMDAO does but I don't know if install locally with `.` together)

nwu63 (dismissed):

nwu63 (commented):
Actually, I would prefer if the tests have more descriptive names (along with the ref files), but I guess that's more of a "nice to have".

bbrelje:
> Actually, I would prefer if the tests have more descriptive names (along with the ref files), but I guess that's more of a "nice to have".

Let's pull it in for now, I need to move on to some other tasks. Right now the test names match the DVGeo tests which are numbered sequentially

nwu63:
Yeah fair, I'm not actually a maintainer for pygeo so I'll wait for others to review.

anilyil (approved):

joanibal (approved):

joanibal (approved):

joanibal:
Sorry @anilyil , @nwu63 
I removed the mpi4py requirement in the setup.py file during the merge and had to put it back. 
Unfortunately that caused your reviews to be dismissed.

nwu63 (approved):

joanibal:
:champagne: Thanks @bbrelje for your hard work on this

PR #52, opened 2020-10-30
Triangulated surface constraint using geograd and multiple DVGeos per DVCon

bbrelje:
## Purpose
This PR adds the KS-based triangulated surface constraint from the Brelje et al AIAAJ paper

## Type of change
What types of change is it?
- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
Reg tests with good coverage have been added. All previous DVConstraints tests pass after the refactor.

## Checklist


pygeo/DVConstraints.py
```diff
@@ -1207,11 +1353,198 @@ def addThicknessToChordConstraints1D(self, ptList, nCon, axis, chordDir,
...
+    def addTriangulatedSurfaceConstraint(self, surface_1_name=None, DVGeo_1_name=None,
+                                         surface_2_name='default', DVGeo_2_name='default',
+                                         rho=50., heuristic_dist=None, perim_scale=0.1, 
+                                         max_perim=3.0, name=None, scale=1., addToPyOpt=True):
+        """
+        Add a single triangulated surface constraint to an aerosurface.
+        This constraint is designed to keep a general 'blob' of watertight
+        geometry contained within an aerodynamic hull (e.g., a wing)
+
+        Parameters
```
joanibal, line 1368: DVGeo_1_name and DVGeo_2_name have no doc strings

pygeo/DVConstraints.py
```diff
@@ -3133,6 +3484,336 @@ def writeTecplot(self, handle):
...
+        # get the point sets
+        self.surface_1_name = surface_1_name
+        self.surface_2_name = surface_2_name        
+        if DVGeo1 is None and DVGeo2 is None:
+            raise UserError('Must include at least one geometric parametrization in constraint '+str(name))
+        self.DVGeo1 = DVGeo1
+        self.DVGeo2 = DVGeo2
+
+        self.surf1_size = surface_1[0].shape[0]
+        # todo double check that these are in proper fortran array order
```
joanibal, line 3505: Has this been done?

pygeo/DVConstraints.py
```diff
@@ -3133,6 +3484,336 @@ def writeTecplot(self, handle):
...
+    def evalTriangulatedSurfConstraint(self):
+        """
+        Call geograd to compute the KS function and intersection length
+        """
+        # first compute the length of the intersection surface between the object and surf mesh
+        from geograd import geograd_parallel
+        mindist_tmp = 0.0
+
+        # first run to get the minimum distance
+        foo, perim_length, mindist, foo2, foo3 = geograd_parallel.compute(self.surf1_p0, self.surf1_p1, self.surf1_p2,
```
joanibal, line 3658: replace foo with `_` if you don't want to save that return value

pygeo/DVConstraints.py
```diff
@@ -3133,6 +3484,336 @@ def writeTecplot(self, handle):
...
+                self.surf2_p0, self.surf2_p1, self.surf2_p2, mindist_tmp, self.rho, self.maxdim)
+        # second run gets the well-conditioned KS
```
joanibal, line 3660: I don't follow. Why does the KS value change between runs?

joanibal: Should these be separate functions?

pygeo/DVConstraints.py
```diff
@@ -3133,6 +3484,336 @@ def writeTecplot(self, handle):
...
+        self.perim_length = perim_length
+        self.minimum_distance = mindist
+
+        if self.perim_length > self.max_perim:
+            failflag = True
+            if mpi4py.COMM_WORLD.rank == 0:
+                print('Intersection length ', str(perim_length), ' exceeds tol, returning fail flag')
+        else:
+            failflag = False
+        # TODO should the perim scale be applied here or elsewhere
```
joanibal, line 3673: Has this been done?

pygeo/DVConstraints.py
```diff
@@ -3133,6 +3484,336 @@ def writeTecplot(self, handle):
...
+
+        Parameters
+        ----------
+        funcsSens : dict
+            Dictionary to place function values
+        """
+        tmpTotal = {}
+
+        if MPI.COMM_WORLD.rank == 0 or not self.mpi:
+            # assume evalFunctions was called just prior and grad was stashed on rank=0
```
joanibal, line 3792: Can you add this requirement to the method doc string.

pygeo/DVConstraints.py
```diff
@@ -3133,6 +3484,336 @@ def writeTecplot(self, handle):
...
+        if self.mpi:
+            volume = MPI.COMM_WORLD.bcast(volume, root=0)
+        funcs[self.name] = volume
+
+    def evalFunctionsSens(self, funcsSens, config):
+        """
+        Evaluate the sensitivity of the functions this object has and
+        place in the funcsSens dictionary
+
+        Parameters
```
joanibal, line 3784: config doc string missing

pygeo/DVConstraints.py
```diff
@@ -3133,6 +3484,336 @@ def writeTecplot(self, handle):
...
+        self.vol_0 = None
+
+        self.nCon = 1
+        return
+
+    def evalFunctions(self, funcs, config):
+        """
+        Evaluate the function this object has and place in the funcs dictionary
+
+        Parameters
```
joanibal, line 3739: config doc string missing

pygeo/DVConstraints.py
```diff
@@ -3133,6 +3484,336 @@ def writeTecplot(self, handle):
         for i in range(len(self.coords)//2):
             handle.write('%d %d\n'% (2*i+1, 2*i+2))
 
+class TriangulatedSurfaceConstraint(GeometricConstraint):
+    """
+    This class is used to enclose a triangulated object inside an
+    aerodynamic surface.
+    """
+
+    def __init__(self, name, surface_1, surface_1_name, DVGeo1, surface_2, surface_2_name, DVGeo2, scale, addToPyOpt,
```
joanibal, line 3493: Can you  add doc strings for the inputs

pygeo/DVConstraints.py
```diff
@@ -3133,6 +3484,336 @@ def writeTecplot(self, handle):
...
+        else:
+            varnamelist = self.DVGeo2.getVarNames()
+        
+        return varnamelist
+
+    def evalFunctions(self, funcs, config):
+        """
+        Evaluate the function this object has and place in the funcs dictionary
+
+        Parameters
```
joanibal, line 3561: config doc string missing

pygeo/DVConstraints.py
```diff
@@ -3133,6 +3484,336 @@ def writeTecplot(self, handle):
...
+        if failflag:
+            funcs['fail'] = failflag
+
+
+    def evalFunctionsSens(self, funcsSens, config):
+        """
+        Evaluate the sensitivity of the functions this object has and
+        place in the funcsSens dictionary
+
+        Parameters
```
joanibal, line 3594: config doc string missing

pygeo/DVConstraints.py
```diff
@@ -3133,6 +3484,336 @@ def writeTecplot(self, handle):
...
+                                wrt=self.getVarNames())
+
+
+
+class TriangulatedVolumeConstraint(GeometricConstraint):
+    """
+    This class is used to compute a volume constraint based on triangulated surface mesh geometry
+    """
+
+    def __init__(self, name, surface, surface_name, lower, upper, scaled, scale, DVGeo, addToPyOpt,
```
joanibal, line 3708: Can you add a doc string for this method

pygeo/DVConstraints.py
```diff
@@ -3897,6 +4578,7 @@ def __init__(self, name, axis, origin, p0, v1, v2, lower, upper, scale,
 
         # Now embed the coordinates and origin into DVGeo
         # with the name provided:
+        # TODO this is duplicating a DVGeo pointset (same as the surface which originally created the constraint)
```
joanibal, line 4581: If this is not done please add an issue on github and reference it in the comment

joanibal (changes requested):
My biggest question is why does geograd have to be called twice? 

most other comments are about doc strings or error handling.

pygeo/DVConstraints.py
```diff
@@ -221,21 +237,52 @@ def setSurface(self, surf):
...
+            if type(surf) == list:
+                # Data from ADflow
+                p0 = numpy.array(surf[0])
+                v1 = numpy.array(surf[1])
+                v2 = numpy.array(surf[2])
+            elif isinstance(surf, str):
+                # Load the surf as a plot3d file
+                p0, v1, v2 = self._readPlot3DSurfFile(surf)
+
+            else: # Assume it's a pyGeo surface
```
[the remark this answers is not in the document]
bbrelje: @joanibal the reason it has to be called twice is that the KS function needs to know the actual minimum distance a priori so it can subtract that off and make the "real" KS computation better conditioned. the same code computes both which is why it gets called twice

pygeo/DVConstraints.py
bbrelje: Yes it has

pygeo/DVConstraints.py
bbrelje: It changes because to obtain a well-conditioned computation of the KS function you subtract off the current minimum value. In exact math it doesn't matter but with computer math it does. So you feed the min dist from the prior run into the second run to get the "real
 KS value

pygeo/DVConstraints.py
bbrelje: The init method is thinly wrapped by the addTriangulatedSurfaceConstraint method which has a comprehensive docstring. The other class __init__ methods in DVConstraints.py don't have docstrings. OK to skip?

pygeo/DVConstraints.py
joanibal: Ok that makes sense. 

Could these two functional calls be combined within geograd so it never returns and ill-conditioned KS?

joanibal: Returning values that aren't accurate or shouldn't be used is a bad idea in my opinion.

@anilyil  thoughts?

bbrelje: @joanibal combining the calls isn't practical because the number of pairwise comparisons is too large to hold in memory pending the minimum being found. geograd isn't intended to be a public API so IMO the way it is now is OK practice (I edited the foos to underscores to make the intent clear)

joanibal: OKk, I suppose that's fine

joanibal (dismissed):

bbrelje:
Don't merge it just yet. I updated geograd with the ability to pass comm from Python layer so I need to update DVConstraints.py real quick

joanibal (dismissed):

pygeo/DVGeometry.py
```diff
@@ -586,6 +589,10 @@ def addGeoDVGlobal(self, dvName, value, func, lower=None, upper=None,
             configurations. The default value of None implies that the design
             variable appies to *ALL* configurations.
         """
+        # if the parent DVGeometry object has a name attribute, prepend it
+        if self.name is not None:
+            dvName = self.name + '_' + dvName
```
anilyil, line 594: how strictly do you want this naming convention of compName+dvName? I was doing compName:dvName in dvgeomulti, but I can also change my approach to be consistent. In my implementation, it was dvgeomulti itself tracking the names I believe.

anilyil:
This looks good in general. One discussion; you are currently allowing multiple DVGeos kicking around with a single DVCon. I was doing this in a slightly different way with DVGeoMulti, where I had an upper-level dvgeo object, that kept track of all of the individual DVgeos for each component. that way we can also define intersections etc. I think we can come up with a common way to handle multiple dvgeos and use a consistent approach.

Here is what I recommend: I understand you don't want to make changes at this point. I also do not want to work on geometry stuff. BUT, I think we can come up with a much better scheme if we sit down together and do a mini hackathon with you. I am guessing we can take care of this work within a few hours on a call. 

What do you say @bbrelje? If you don't want to make changes, that is totally fine by me and the PR is good as far as I am concerned.

anilyil:
I talked to ben and we decided to merge this PR now, and we will refactor these stuff for harmony when we merge the dvgeomulti code. We estimate to do this in April, when ben will be done running the cases for his thesis work. I will create an issue about this and also move my dvgeomulti code to my public fork because the paper is accepted and in press now.

anilyil (dismissed):

anilyil:
I have created the issue that details the refactoring in #57.

bbrelje:
I forgot I never included the triangulated volume constraint when I ported from the tensorflow-based geograd to the fortran one. I will need to deprecate that function or add it to geograd this weekend (probably the former). Block merge for now.

anilyil:
@bbrelje I converted the PR to a draft. I understand you want to work on this more? I did the draft thing to prevent somebody accidentally merging this since both reviewers approved.

anilyil (approved):

joanibal (approved):

nwu63:
`geograd` tests failed previously, this has now been fixed. I've verified locally that all tests pass, so I'll merge manually.

PR #55, opened 2020-12-21
added child warning example

joanibal:
## Purpose
Added an example test to show some non-intuitive child FFD behavior 

## Type of change
What types of change is it?

- [x] Other (please describe)

## Testing
Explain the steps needed to test the new code to verify that it does indeed address the issue and produce the expected behavior.

## Checklist


bbrelje (changes requested):
@joanibal, first of all, I am sorry for the delay. It's not excusable. 

Are you intending this example to be part of the reg test suite? I checked the logs and I didn't see that test_parent_shape_child_rot() is being run. I think that the warning_childFFD.py needs to be renamed to begin with the word test_ in order for pytest to pick it up.

anilyil:
The changes look good. I would say we need to document this far better. I only know about this issue because we discussed it in the office. I think we should include an in depth explanation of this issue in the code. I dont remember the details 100% but I can help with this if you refresh my memory a bit. We should also put in comments in the childffd code pointing to this test and this behavior, so people who want to use that feature will be aware of this.

joanibal:
@bbrelje  I didn't intend it to be run from the reg test suite.

joanibal:
@anilyil 
I agree, I added some detail in issue #7.
Do you think we should move this info else where?

bbrelje (dismissed):

bbrelje:
I'm good with this one, then. Never used the child FFD feature so I didn't understand the backstory here.

anilyil:
@joanibal Can we refer to issue #7 in the code also? Finally, I think the figures on the issue can be improved. They are difficult to understand because too many things going on. The cube example you used to explain the issue to me in FXB was great. I am guessing the script in this PR also uses that example. Would it be too much work to add those figures to the issue?

joanibal:
@anilyil 
I added a reference to the issue in the code. 

I will update the issue later to use a simpler example.

anilyil (approved):

bbrelje (approved):

PR #60, opened 2021-01-25
Fixed all the complex to real assignment warnings from pygeo.

### LINKED ISSUE #31: Complex cast in pyNetwork
## Type of issue
What types of issue is it?
_Select the appropriate type(s) that describe this issue_

- Bugfix (non-breaking change which fixes an issue)

## Description
Running any ASO will cause the following warning:

```
$HOME/mdolab/pygeo/pyNetwork.py:147: ComplexWarning: Casting complex values to real discards the imaginary part
  self.curves[icurve].coef[i] = self.coef[ii]
```

To reproduce, run then ASO script in MACH-Aero-tutorial.

anilyil:
## Purpose

This PR fixes all of the complex to real assignments that come up with pyGeo. I ran all of the pyGeo tests and fixed all of the warnings that showed up. So reviewers, please let me know if you find other instances of warnings and I can update this PR. I also ran the adflow complex tests. There is at least one that uses complex DVGeo and it passes.

pyGeo uses complex step internally to compute the sensitivities when run in real mode. Due to this, there may be some internal inconsistencies where real values were assigned complex values. I think all of the warnings I caught, the issue was caused by two things:
1. Conversion from complex to real was wrong. The code used `array.astype('d')` to convert to real but this raises a warning. I converted these to `array.real.astype('d')`. This likely has no effect on the actual result; the code was trying to use just the real part to begin with.
2. Some internal coefficients were left in real data type during a complex-step process. To fix these, I modified the order of calls in DVGeo so that these underlying coefficients are complexified correctly before the actual complex values started propagating. This was a potential bug previously in the code. The reg_tests still pass with my changes, so I am not sure if this was a bug, or was not covered. The only change that fell into this group was around line 2924 in DVGeo: https://github.com/mdolab/pygeo/pull/60/files#diff-a4f2a97516d34a341280073c82fc158317f3781e0eb397c4cbaa72d698912a71R2924 

## Type of change

- [x] Maintenance update

## Testing

No new tests. The warnings are gone.

## Checklist

Closes #31.

### DIFF
--- a/pygeo/DVConstraints.py
+++ b/pygeo/DVConstraints.py
@@ -3503,7 +3503,7 @@ class LinearConstraint(object):
             elif key in self.DVGeo.DV_listSectionLocal:
                 cons.extend(self.jac[key].dot(self.DVGeo.DV_listSectionLocal[key].value))
 
-        funcs[self.name] = numpy.array(cons).astype('d')
+        funcs[self.name] = numpy.array(cons).real.astype('d')
 
     def evalFunctionsSens(self, funcsSens):
         """
@@ -5344,7 +5344,7 @@ class GlobalLinearConstraint(object):
         for key in self.jac:
             cons.extend(self.jac[key].dot(self.DVGeo.DV_listGlobal[key].value))
 
-        funcs[self.name] = numpy.array(cons).astype('d')
+        funcs[self.name] = numpy.array(cons).real.astype('d')
 
     def evalFunctionsSens(self, funcsSens):
         """
--- a/pygeo/DVGeometry.py
+++ b/pygeo/DVGeometry.py
@@ -1157,8 +1157,10 @@ class DVGeometry(object):
             # **Important**: this expects the FFD coef to be clean on this level,
             # meaning that the only changes to FFD.coef can be coming from
             # higher levels.
-            if isComplex:
-                self.links_x = self.links_x.astype('D')
+
+            # just use complex dtype here. we will convert to real in the end
+            self.links_x = self.links_x.astype('D')
+
             for ipt in range(self.nPtAttach):
                 base_pt = self.refAxis.curves[self.curveIDs[ipt]](self.links_s[ipt])
                 self.links_x[ipt] = self.FFD.coef[self.ptAttachInd[ipt], :] - base_pt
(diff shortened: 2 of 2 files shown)


anilyil:
both complex and real adflow tests ran successfully w/o warnings. I also ran all of the pygeo tests and they also finish w/o warnings

anilyil:
The changes around https://github.com/mdolab/pygeo/pull/60/files#diff-a4f2a97516d34a341280073c82fc158317f3781e0eb397c4cbaa72d698912a71R2924 turned out to be not a bug. Here is an explanation:

That part is inside a complex-step jacobian computation. The code perturbs the DV and updates the reference axis coordinates. The call that updates the ref.axis coordinates is `_updateCurveCoef` from pyNetwork. This call: https://github.com/mdolab/pygeo/blob/master/pygeo/DVGeometry.py#L2922 was trying to set the perturbed coefficients in pyNetwork but because the `_complexifyCoef` was done after this, the coefficients were not complexified. These coefficients were supposed to receive these complex-perturbed values but before this PR, they would only take in the real part as the warning prints out.

Luckily, in the following `_update_deriv` call, there is another call to `_updateCurveCoef`, which happens after `_complexifyCoef` was called, so even though the code was not correct, this did not result in any wrong values. 

The PR fixes this issue; the `_complexifyCoef` is called before `_updateCurveCoef` and the code can now put the correct complex-perturbed coefficients in the curves. I have also checked the actual values in some tests that use this part of the code, and the actual values of the derivatives is unaffected by the change. The PR removes the warnings, and fixes the incorrect code. The issue in the code did not affect the final results. (this is all based on my quick read of the code so I can be very wrong).

bbrelje (approved):
I have to admit I have no understanding of the math that's going on here without diving deep into many kLOC  but these changes look defensible to me and the pygeo reg tests and adflow integration tests definitely cover the affected methods.

bbrelje:
If there is somebody else on the team with a better understanding of the FFD math underlying the code it might be good to get review from them but to the best of my ability this looks good.

nwu63 (approved):

nwu63:
LGTM, works as advertised. @joanibal do you have time to take a look? Otherwise I can merge.

PR #72, opened 2021-03-09
pyGeo Geometry Deformation given DVGeometry FFD

bernardopacini:
## Purpose
This PR adds a function within DVGeometry to update a pyGeo geometry object with a given FFD deformation. The purpose of this function is to allow users to generate an IGES (or Tecplot) file of their updated geometry to be used after the optimization (for example when using CAD). An example usage of this function is included in `./examples/deform_geometry/` and associated documentation is added in the pyGeo documentation under `Tutorials`.

To allow for a tutorials / examples section in the docs, I reorganized the documentation structure to have separate categories: installation, tutorials, and API.

## Type of change
What types of change is it?

- [X] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
This function is separate from current tests and is difficult to test as it's output is an IGES or Tecplot file.

## Checklist

- [X] I have added necessary documentation


marcomangano:
Well done! Looking forward to check the docs.
Just one question: are the FFD and wing geometry scripts you added a duplicate of what is in MACH-Aero tutorial?

bernardopacini:
Yes, they are. I was debating removing them and just leaving the files (.dat, .igs, .xyz) to avoid duplication, but I did not want to leave a loose end in the future. What do you think?

examples/deform_geometry/runScript.py
```diff
@@ -0,0 +1,163 @@
...
+    # add the twist design variable to DVGeo
+    DVGeo.addGeoDVGlobal(dvName="twist", value=twist0, func=twist, lower=-10, upper=10, scale=1.0)
+
+    # =========================================================================
+    # Update pyGeo Object and output result
+    # =========================================================================
+    DVGeo.updatePyGeo(geo, "tecplot", "wingNew", nRefU=10, nRefV=10)
+
+if __name__ == "__main__":
+    if len(sys.argv) != 1:
```
nwu63, line 153: We generally prefer `argparse` over raw `sys.argv` parsing.

pygeo/DVGeometry.py
```diff
@@ -2133,6 +2133,81 @@ def writePlot3d(self, fileName):
...
+            Number of spline refinement points to add in the surface B-Spline u-direction.
+            If scalar, it is applied across each surface. If list, the length must match the
+            number of surfaces in the object and corresponding entries are matched with surfaces.
+        nRefV: int or list of ints
+            Number of spline refinement points to add in the surface B-Spline v-direction.
+            If scalar, it is applied across each surface. If list, the length must match the
+            number of surfaces in the object and corresponding entries are matched with surfaces
+        """
+        # Refine Surface -- U-Direction
+        if type(nRefU) is int:
```
nwu63, line 2159: Type checking should be done using `isinstance()` rather than `type()`

nwu63 (commented):
Just some minor code reviews. Make sure @bbrelje takes a look at this since he is also working on some docs, so there might be conflicts.

examples/deform_geometry/runScript.py
bernardopacini: Fixed.

pygeo/DVGeometry.py
bernardopacini: Fixed.

pygeo/DVGeometry.py
```diff
@@ -2133,6 +2133,81 @@ def writePlot3d(self, fileName):
...
+                for iY in Refine_V:
+                    geo.surfs[iSurf].insertKnot('v', iY, 1) 
+        else:
+            raise TypeError("nRefV type not recognized, must be: integer or list of integers")
+
+        # Update Coefficients
+        for iSurf in range(geo.nSurf):
+            # Add Point Sets
+            npt = geo.surfs[iSurf].nCtlu*geo.surfs[iSurf].nCtlv
+            self.addPointSet(geo.surfs[iSurf].coef.reshape((npt, 3)), 'coef%d'%iSurf)
```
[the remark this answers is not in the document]
bernardopacini: I double checked the `addPointSet()` function [here](https://github.com/mdolab/pygeo/blob/17cb3fcf49497f185bad118d39849bf91c2753e7/pygeo/DVGeometry.py#L454) as well as the FFD generation function [here](https://github.com/mdolab/pygeo/blob/17cb3fcf49497f185bad118d39849bf91c2753e7/pygeo/pyBlock.py#L78), and from what I can tell the process is as follows:

1) The point set is added only in name, essentially initializing the arrays to be used later. The points are converted into the required data type and added to the DVGeometry object, but the FFD coefficients are not computed:
```
self.ptSetNames.append(ptName)
self.zeroJacobians([ptName])
self.nPts[ptName]=None

points = numpy.array(points).real.astype('d')
self.points[ptName] = points
```

2) The FFD is *reset* to its original shape, using `self.FFD.coef = self.origFFDCoef`:
```
if origConfig:
    tmpCoef = self.FFD.coef.copy()
    self.FFD.coef = self.origFFDCoef
    self.FFD._updateVolumeCoef()
```

3) The points are attached to the now undeformed FFD:
```
if self.isChild:
    self.FFD.attachPoints(
        self.points[ptName], ptName, interiorOnly=True, **kwargs)
else:
    self.FFD.attachPoints(
        self.points[ptName], ptName, interiorOnly=False)
```

4) The FFD is updated back to its deformed shape:
```
if origConfig:
    self.FFD.coef = tmpCoef
    self.FFD._updateVolumeCoef()
```

5) The derivatives of the points are computed and the coefficients are *not* updated:
```
self.FFD.calcdPtdCoef(ptName)
self.updated[ptName] = False
```

This all hinges on the `origConfig` argument being set to `True` for `addPointSet()`, but it is not possible to set that to `False` in my added function.

This seems to be the expected behavior, but do you remember what exactly was tricky about adding them? I don't want to miss something in what is going on.

examples/deform_geometry/runScript.py
nwu63: This can actually be improved:
```
parser.add_argument("--output_type", type=str, default="iges", choices=["iges", "plot3d", "liftingsurface"])
```
Then you don't need the `else` block.

bernardopacini: Thanks for the recommendation -- updated.

nwu63: What I meant was that because you specified `choices`, if the value given by the user does not fall into those, then `argparser` itself will [raise the error](https://docs.python.org/3/library/argparse.html#choices). So there is no need for the `else` block altogether since any error would be raised at the `parse_args()` call already.

bernardopacini: Yes, sorry I forgot to delete that line after updating the parsing line. I fix that.

nwu63 (approved):

nwu63:
Not sure why the docs build is not triggering, I can just manually merge this if another maintainer is able to review the PR.

anilyil:
@bbrelje do you need more changes in this PR? @bernardopacini do you want me to review this?

bernardopacini:
If you have time, that would be great @anilyil.

bernardopacini:
I am seeing some strange behavior in DAFoam when I try to use this. I do not think it is related, but just to be sure let's hold off on merging this until I have that working properly.

joanibal:
It looks like this PR adds a couple of large files 
examples/deform_geometry/geo/wing.dat (650,000+ lines)
examples/deform_geometry/geo/wing.igs (98,000+ lines)

Should we download those like we do the mesh files for adflow?

nwu63:
> It looks like this PR adds a couple of large files
> examples/deform_geometry/geo/wing.dat (650,000+ lines)
> examples/deform_geometry/geo/wing.igs (98,000+ lines)
> 
> Should we download those like we do the mesh files for adflow?

Actually that's a good point. However, once it's committed it's pretty difficult to purge since if you just remove the files they're still in the history (and will get downloaded when you run `git fetch`). If we really want to remove those you may have to do some rebasing/force pushes.

bernardopacini:
The issue was with the DAFoam case, so this is ready after all.

Would you like me to try to remove the large files from the history?

nwu63:
Do you know how you would go about doing it? I'm not 100% sure if rebasing + force push would work, or if you need to make a new branch and cherrypick.

anilyil:
Doesn't squash merging just do one commit with the files that are remaining in the last version of the PR? So I think we can just remove the files and once we squash merge, it will only add the changed files in the git history of mdolab/pygeo. The files will stay in the history of @bernardopacini's fork though

examples/deform_geometry/runScript.py
```diff
@@ -0,0 +1,161 @@
...
+        tip="rounded",
+        bluntTe=True,
+        squareTeTip=True,
+        teHeight=0.25 * 0.0254,
+    )
+    # rst LiftingSurface (end)
+
+    # =========================================================================
+    # Setup DVGeometry object
+    # =========================================================================
```
anilyil, line 55: Is the dvgeo part identical across these scripts? if so, we can just have another smaller function that creates the DVGeo object and just returns it? Less code duplication that way I think?

examples/deform_geometry/runScript.py
```diff
@@ -0,0 +1,161 @@
...
+    def twist(val, geo):
+        for i in range(nRefAxPts):
+            geo.rot_z["wing"].coef[i] = val[i]
+
+    # set initial twist, we set the root twist to 5 deg
+    twist0 = [0] * nRefAxPts
+    twist0[0] = 5
+
+    # add the twist design variable to DVGeo
+    DVGeo.addGeoDVGlobal(dvName="twist", value=twist0, func=twist, lower=-10, upper=10, scale=1.0)
```
anilyil, line 72: I see that you are adding the twist dv and setting its value. I think this is not how people are used to work with DVGeo. It might be better to modify this line to just add the twist, and then add a few more lines where you call `setDesignVars` with the new design variables. This way, I think it will be easier for people to just extract the dictionaries from their optimizations and paste it into this code.

pygeo/DVGeometry.py
```diff
@@ -2133,6 +2133,81 @@ def writePlot3d(self, fileName):
...
+            number of surfaces in the object and corresponding entries are matched with surfaces.
+        nRefV: int or list of ints
+            Number of spline refinement points to add in the surface B-Spline v-direction.
+            If scalar, it is applied across each surface. If list, the length must match the
+            number of surfaces in the object and corresponding entries are matched with surfaces
+        """
+        # Refine Surface -- U-Direction
+        if isinstance(nRefU, int):
+            # Refine BSplines by adding knot points
+            Refine_U = numpy.linspace(0.0, 1.0, nRefU)
```
anilyil, line 2161: I am not sure how knots are added in pygeo. What would happen if one of the u locations correspond to an existing knot? I think if there is no check for this, it will change the continuity of the bspline at that knot. Do we check this? I am not necessarily asking for changes here, just trying to understand.

Also, my bspline math is a bit rusty, but I think the u=0 and 1 positions will have as many knots as the order of the bspline (or something similar?). Does adding more points by default make sense there? Maybe we can just add 2 to `nRefU`, get the linspace, and then dont add the knots at u=0 and 1 locations.

anilyil (changes requested):
I added a few comments. Besides that the changes and the documentation looks good.

nwu63:
> Doesn't squash merging just do one commit with the files that are remaining in the last version of the PR? So I think we can just remove the files and once we squash merge, it will only add the changed files in the git history of mdolab/pygeo. The files will stay in the history of @bernardopacini's fork though

I may be wrong, but I believe the individual commits are still transferred also, it will just appear as a single commit. For example, individual commits like [this](https://github.com/mdolab/pygeo/pull/73/commits/8fb2f0e1ebde32f49836e296c84aa3bc37b85d96) are still accessible after it's squashed and merged.

anilyil:
oh I see, so they would take up space in the mdolab repo right?

anilyil:
okay that makes sense, then I will also need to do that for the overset tutorial. Thanks for the explanation @nwu63!

nwu63:
Yeah unfortunately purging large files from a repository is not easy. I tried to do that for ADflow once and nearly screwed everything up.

bernardopacini:
That makes sense, I can go through that process once the PR is approved. I'll address the other comments first.

Another option might be using the [git filter-branch](https://docs.github.com/en/github/authenticating-to-github/removing-sensitive-data-from-a-repository) command, but I have not used it before so I am not sure how exactly it will work.

examples/deform_geometry/runScript.py
bernardopacini: Yes, they are the same. I thought about combining them but figured it would be easier for a user to see the complete example. I see your point though, I'll convert it into a function to avoid repeating those lines.

examples/deform_geometry/runScript.py
bernardopacini: Good point, I'll change this to be updated using the `setDesignVars()` dictionary function.

nwu63:
Yeah actually if you wanna try that it sounds like a better solution. We can talk offline to work through that after you address the rest of the comments.

pygeo/DVGeometry.py
bernardopacini: For your second point, knot points are automatically *not* added at the ends of the spline where `u = 0 or 1` (there is a check for this in pySpline [here](https://github.com/mdolab/pyspline/blob/a3b45e9aad5938b09a318b5f570bd0335dabfc55/pyspline/pySpline.py#L1795)). You bring up a good point that `nRefU = 0 or 1` cause no refinement (for different reasons), so I will offset that to refine properly.

For your first point, I am actually not sure either. From what I can tell from the [`insertKnot.f90`](https://github.com/mdolab/pyspline/blob/master/src/insertKnot.f90) file, there is a check for multiplicity:

```
! Compute its multiplicity
s = 0 ! Knot multiplicity
mult: do i=0,k-1
   if (abs(t(ileft-i) -u ) < 1e-12) then 
      s = s + 1
   else
      exit mult
   end if
end do mult

! We need to make sure that the requested multipliity r, plus 
! the actual multiplicity of this know is less than (k-1)

if (s + r + 1 > k) then
   r = k-1-s
end if

! If we *can't* insert the knot, we MUST copy t and coef to t_new
! and coef_new and return
if (r == 0) then
   coef_new(:,1:nctl) = coef(:,1:nctl)
   t_new(1:nctl+k) = t(1:nctl+k)
   return
end if
```

My knowledge of bspline math is limited, so I am not entirely sure what the process here is. But, it seems that if there is multiplicity and it will cause the number of knot points to exceed the order at that point, the inserted point is simply not added. This may cause unexpected behavior for the user if their geometry is refined less than expected, but realistically if their geometry already has a point where it is being refined it should not make a big difference. I can add a warning about this in the docs.

Is this an issue we want to open for pySpline?

anilyil: I see, I think pyspline is fine, but we should _not_ add the knot if there is another knot there. Changing the multiplicity from 1 to 2 would change the continuity of the derivatives of the bspline, and instead of refining it, it will change the curve itself. Is there an easy way we can get the current knot vector to check we are only refining and not modifying the multiplicity anywhere?

bernardopacini: I will take a look to see if this is possible.

examples/deform_geometry/runScript.py
anilyil: The way I had this in mind was that we would have a function that would return the built DVGeo object, but I think your version might be easier to follow since the flow is linear. So I think this looks good!

pygeo/DVGeometry.py
anilyil: Sounds great. The addition of 2 makes sense if pyspline is not adding the knots at 0 and 1 by default. Let me know if this is possible, if it takes too much work, you can also just add a few comments explaining this behavior so that people are forced to understand what the bspline knots are doing? 

Anyways, besides this the PR looks good for me.

bernardopacini: It is hard coded. I'll hard code it here too.

anilyil: Yeah then hard coding is just fine. Leave a comment in the code that either points here or explains why 1e-12 is there, and we should be set. (Out of curiosity, how does the knot vector look like for your example? Does it actually have values that might be overlapping or was my suggestion a complete non-issue?)

bernardopacini: I updated it and added an internal function to check the tolerance. Let me know if it makes sense.

anilyil: Neat, the check looks great. I am not sure how big of a difference you would get; it depends on the order of the spline and the actual positions of the control points etc etc but I think doing it this way should be more failsafe. The rest of the PR looks good to me!

anilyil (dismissed):

marcomangano:
> Yeah unfortunately they would. If you want to be sure that they are removed, I would recommend the following once the rest of the code is approved:
> 
>     * close this PR
> 
>     * make a new branch from `master`
> 
>     * set the state of this branch to be the same as this PR branch (you may have to have another directory and copy/paste, I don't know how to check out the state of a branch into the working space)
> 
>     * add everything except for the files we don't want to commit
> 
>     * make a PR new PR and link to this PR so we preserve the discussion
> 
> 
> We lose the commit history and the discussions are going to be in another PR, but this will work. Alternatively, rebasing/force pushing might work too but I'm not 100% sure.

@bernardopacini you could also use the same approach... but replicate the steps commit by commit, so to not lose the history:
1. Make a clean copy of pyGeo and set a new branch
2. On the current branch (and folder), roll back to the [first commit of this PR](https://github.com/mdolab/pygeo/pull/72/commits/8dcd6b1ebbee80f6894d725f7915485556c18ea2)
3. Copy the files from your active pyGeo folder to the new pyGeo folder, and recommit with the same message
4. Back to the current branch, move one commit ahead, then copy to the other folder and recommit
5. Continue this until the last commit here, this time avoiding to add unwanted files.

This way we don't use the history. Does it make sense @nwu63 ? Also happy to help offline.

nwu63:
I don't think this is entirely necessary since this is a smaller development, but you can do it if you want I guess.

marcomangano:
Sure, I guess this is really up to the developer, depending on how much the single steps are relevant for future maintenance

bernardopacini:
This PR is approved and I moved the large files out of the current state (downloadable via the `get-input-files.sh` script instead). I will now try to remove the large files from the history before we merge this.

nwu63:
After some testing with @bernardopacini, we believe that because this PR comes from a fork, squash + merge will not add the large files to the `mdolab` repo. Since @anilyil has approved already I'm going to merge this.

nwu63 (approved):

PR #79, opened 2021-04-24
Added tests for example scripts

nwu63:
## Purpose
Similar to some other repos, I have added tests for the example scripts. This will ensure that any changes in API will be reflected in these scripts, and perhaps will increase our testing coverage slightly.

## Type of change
What types of change is it?

- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Testing
I added new tests.

## Checklist


marcomangano (dismissed):
Looks good, thanks for working on this!
I assume that accumulating the output files in the `self.output_file_list` attribute is necessary for testflo to use the `tearDown` method at the end of the run, right?

nwu63:
> I assume that accumulating the output files in the `self.output_file_list` attribute is necessary for testflo to use the `tearDown` method at the end of the run, right?

Yeah this approach requires `self.output_file_list` to be set in the test method. Perhaps it would be a good idea to set this to an empty list in the `setUp` function?

marcomangano (approved):

joanibal:
it looks like input_files/get-input-files.sh is an empty file?
why do we run an empty script? for inter repo consistency?

setup.py
```diff
@@ -1,49 +1,21 @@
...
-        booktitle = {Proceedings of the 13th AIAA/ISSMO Multidisciplinary Analysis Optimization Conference},
-        year = {2010},
-        note = {AIAA 2010-9231},
-        month = {September},
-        address = {Fort Worth,~TX},
-        author = {Gaetan K. W. Kenway and Graeme J. Kennedy and Joaquim R. R. A. Martins}
-      }
-      
-      """,
+    long_description=long_description,
```
joanibal, line 18: nice

joanibal (approved):

joanibal:
Looks good, just one small question

nwu63:
> it looks like input_files/get-input-files.sh is an empty file?
> why do we run an empty script? for inter repo consistency?

Nope the file exists, see [here](https://github.com/mdolab/pygeo/blob/master/input_files/get-input-files.sh). It's shown as an empty diff because the only change I made there was add the execution permission via `chmod +x`. `git` keeps track of (only) this bit of file permissions so when you pull the file it will be automatically set as an executable or not.

joanibal:
Oh, I see. 
Thanks for explaining that

PR #82, opened 2021-05-07
Added spline parameters as kwarg for plot3d-based objects

marcomangano:
## Purpose
The current `_readPlot3D` method in `pyGeo` calls `pySpline` with a hard-coded set of parameters for spline order and number of control points. For some geometries, the object that is returned from `pyGeo` can be very low quality, as for example this wind turbine blade (note the bad interpolation and the crossing surfaces at the trailing edge):

This is not an issue when using `iges` inputs, as this format also provides interpolation information.
To resolve this issue (thanks to @eirikurj's tip), I just added the order of the splines and the number of control points in the u and v directions as `*kwargs` to the  `pyGeo()` class.
By manually setting the parameters - `nCtlu=12, nCtlv=26` in this example- the user can get appropriate interpolations:

I don't think there is a way to test this as:
- We would need the compare the output object with a reference file, not sure `assert_allclose` would be appropriate
- This is actually a pyspline feature, I am just removing the hard-coded values

Looking forward to your review! 

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
There is no real test to this as it is just a matter of calling `pySpline` with different parameters.

## Checklist

### DIFF
--- a/pygeo/pyGeo.py
+++ b/pygeo/pyGeo.py
@@ -116,6 +116,14 @@ class pyGeo:
             File name to load. Should end in .xyz
         order : str
             'f' for fortran ordering (usual), 'c' for c ordering
+        u : int
+            Spline order in u
+        kv : int
+            Spline order in v
+        nCtlu : int
+            Number of control points in u
+        nCtlv : int
+            Number of control points in v
         """
         f = open(fileName, "r")
         binary = False
--- a/pygeo/pyGeo.py
+++ b/pygeo/pyGeo.py
@@ -107,7 +107,7 @@ class pyGeo:
     #               Initialization Type Functions
     # ----------------------------------------------------------------------------
 
-    def _readPlot3D(self, fileName, order="f"):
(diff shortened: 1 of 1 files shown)


marcomangano:
> Makes a lot of sense, you are just promoting the hard-coded values as paramters. Two small points:
> 
>     * Probably makes sense to add the same docstrings for the `__init__` function since that's where the user would actually pass the values
> 
>     * Maybe add intersphinx links to pyspline and link to the relevant section there? I'm not even sure if we have docs on this in pyspline

1. There are no docstrings in the `__init__` function, maybe we should add stuff from all the functions there?
2. Pls teach me how

marcomangano:
My bad about point 1. I am not sure how you want to add the links here though.

nwu63 (approved):
Forget what I said, I think this is okay. In the future, we should absolutely refactor some of this, since we have similar options like `kSpan` and so on for the IGES initialization.

eirikurj (approved):

PR #87, opened 2021-06-24
Removed unneccesary mpi4py import

### LINKED ISSUE #85: Importing mpi4py Unnecessarily when Importing pygeo.geo_utils
## Description
When scripting together a bunch of runs its is useful to be able to call `os.system` or `subprocess.run` on `mpirun` commands. During the setup up for various runs when grids are created `pygeo` cannot currently be used to correct or modify geometries because it will break any calls to command line `mpirun`. I think this is due to pygeo importing all of its modules whenever it is called, which in some cases also imports `mpi4py`.  System calls cannot be made that use `mpirun` if `mpi4py` was previously imported. I am not too familiar with the `pygeo` structure, but there may be other submodules that this sort of thing also applies to.

### Steps to reproduce issue
The steps are pretty much as follows, an example script is seen directly below the steps

1. `import` something from pygeo
2. try to run either `os.system("mpirun hostname")` or `subprocess.run("mpirun hostame", shell=True)`

```
>>> from pygeo.geo_utils import *
>>> import subprocess
>>> print(subprocess.run("mpirun hostname", shell=True))
```

### Current behavior
The current behavior returns a non-zero error code on any system calls with `mpirun` calls. For the attached script this looks like:
```
$ python test.py 
CompletedProcess(args='mpirun hostname', returncode=1)
```

### Expected behavior
For the sample script the expected behavior is a list of all of the processors on your machine. For my 24 core workstation it looks like this:
```
$ python test.py 
akleb-workstation
akleb-workstation
akleb-workstation
akleb-workstation
akleb-workstation
akleb-workstation
akleb-workstation
akleb-workstation
akleb-workstation
akleb-workstation
akleb-workstation
akleb-workstation
akleb-workstation
akleb-workstation
akleb-workstation
akleb-workstation
akleb-workstation
akleb-workstation
akleb-workstation
akleb-workstation
akleb-workstation
akleb-workstation
akleb-workstation
akleb-workstation
CompletedProcess(args='mpirun hostname', returncode=0)
```

kanekosh:
## Purpose
Removed the implicit import of `mpi4py` when importing `geo_utils` etc. Closes #85.
Now the example provided by @akleb returns:
```
>>> from pygeo.geo_utils import *
>>> import subprocess
>>> print(subprocess.run("mpirun hostname", shell=True))
bdeb688de7b2
bdeb688de7b2
bdeb688de7b2
bdeb688de7b2
CompletedProcess(args='mpirun hostname', returncode=0)
```

## Backwards imcompatibility
`DVGeometryVSP` import.
**previous:** `from pygeo import DVGeometryVSP`, which no longer works.
**new:** `from pygeo.DVGeometryVSP import DVGeometryVSP`
This will affect a few files in ADflow and personal runscripts. I'll submit a PR to ADflow.

## Type of change
What types of change is it?

- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/pygeo/DVConstraints.py
+++ b/pygeo/DVConstraints.py
@@ -4,7 +4,6 @@
 import numpy as np
 from . import geo_utils, pyGeo
 from pyspline import pySpline
-from mpi4py import MPI
 from scipy.sparse import csr_matrix
 from collections import OrderedDict
 
@@ -3898,6 +3897,7 @@ class TriangulatedSurfaceConstraint(GeometricConstraint):
         """
         # first compute the length of the intersection surface between the object and surf mesh
         from geograd import geograd_parallel
+        from mpi4py import MPI
 
         mindist_tmp = 0.0
 
@@ -3945,6 +3945,7 @@ class TriangulatedSurfaceConstraint(GeometricConstraint):
         """
         # first compute the length of the intersection surface between the object and surf mesh
         from geograd import geograd_parallel
+        from mpi4py import MPI
 
         deriv_output = geograd_parallel.compute_derivs(
             self.surf1_p0,
@@ -5711,8 +5712,9 @@ class CurvatureConstraint(GeometricConstraint):
         for iSurf in range(self.nSurfs):
             self.curvatureRef += self.evalCurvArea(iSurf)[0]
 
-        if MPI.COMM_WORLD.rank == 0:
-            print("Reference curvature: ", self.curvatureRef)
+        # from mpi4py import MPI
+        # if MPI.COMM_WORLD.rank == 0:
+        #     print("Reference curvature: ", self.curvatureRef)
 
     def evalFunctions(self, funcs, config):
         """
@@ -5828,8 +5830,9 @@ class CurvatureConstraint(GeometricConstraint):
             # Now compute the KS function for mean curvature, equivelent to KS(H*H*dS)
             sigmaH = np.dot(one, np.exp(self.KSCoeff * H * H * dS))
             KSmean = np.log(sigmaH) / self.KSCoeff
-            if MPI.COMM_WORLD.rank == 0:
-                print("Max curvature: ", max(H * H * dS))
+            # from mpi4py import MPI
+            # if MPI.COMM_WORLD.rank == 0:
+            #     print("Max curvature: ", max(H * H * dS))
             return [KSmean, K, H, C]
         else:
             raise Error(
--- a/pygeo/DVGeometry.py
+++ b/pygeo/DVGeometry.py
@@ -5,7 +5,6 @@ import copy
 from collections import OrderedDict
 import numpy as np
 from scipy import sparse
-from mpi4py import MPI
 from pyspline import pySpline
 from . import pyNetwork, pyBlock, geo_utils
 import os
@@ -1953,6 +1952,7 @@ class DVGeometry(object):
                 dIdx_local[i, :] = self.JT[ptSetName].dot(dIdpt[i, :, :].flatten())
 
         if comm:  # If we have a comm, globaly reduce with sum
+            from mpi4py import MPI
             dIdx = comm.allreduce(dIdx_local, op=MPI.SUM)
         else:
             dIdx = dIdx_local
--- a/pygeo/DVGeometry.py
+++ b/pygeo/DVGeometry.py
@@ -1953,7 +1953,7 @@ class DVGeometry(object):
 
         if comm:  # If we have a comm, globaly reduce with sum
             from mpi4py import MPI
-            
+
             dIdx = comm.allreduce(dIdx_local, op=MPI.SUM)
         else:
             dIdx = dIdx_local
--- a/pygeo/DVGeometry.py
+++ b/pygeo/DVGeometry.py
@@ -1953,6 +1953,7 @@ class DVGeometry(object):
 
         if comm:  # If we have a comm, globaly reduce with sum
             from mpi4py import MPI
+            
             dIdx = comm.allreduce(dIdx_local, op=MPI.SUM)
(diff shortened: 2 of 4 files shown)


pygeo/DVConstraints.py
```diff
@@ -5711,8 +5712,9 @@ def __init__(self, name, surfs, curvatureType, lower, upper, scaled, scale, KSCo
         for iSurf in range(self.nSurfs):
             self.curvatureRef += self.evalCurvArea(iSurf)[0]
 
-        if MPI.COMM_WORLD.rank == 0:
-            print("Reference curvature: ", self.curvatureRef)
+        # from mpi4py import MPI
+        # if MPI.COMM_WORLD.rank == 0:
+        #     print("Reference curvature: ", self.curvatureRef)
```
marcomangano, line 5717: I think the print should remain... but this could become messy when we use pygeo within an MPI job. We could use `pp` from [baseclasses](https://github.com/mdolab/baseclasses/blob/4697ae6154906f0c1f7d28dfd30319ead9adfd05/baseclasses/BaseSolver.py#L227-L241) but this would add the dependency from `baseclasses`... Opinions? @nwu63

marcomangano (changes requested):
I am fine with the backwards incompatibilities if we all agree with that, just see my point below

pygeo/DVConstraints.py
nwu63: No we shouldn't use baseclasses because that requires rewriting the whole package to use BaseSolver which doesn't make sense to me.

nwu63:
The more I'm thinking about this, I think there are valid scenarios where we want to load MPI, for printing or whatever when we're running this code in parallel. What we essentially want to do is control whether or not to load MPI at run time, and the decision is not simply checking whether `mpi4py` is installed because you may want to not load MPI even if it's available.

If we want to do this, then what we can do is introduce an environment variable which will control the MPI behaviour of the code, similar to what is done in pyOptSparse or OpenMDAO currently (via `PYOPTSPARSE_REQUIRE_MPI` and `OPENMDAO_USE_MPI`, respectively). If we uniformly adopt this across MACH this might help ensure consistency in detecting/loading MPI. Thoughts?

anilyil:
I dont think removing mpi4py is necessary like Neil said. We can use a dummy mpi4py like openmdao if its not available. I dont like the separate import of DVGeoVSP. I also dont like that we import MPI during that eval functions sense call. 

This code is designed to work in a parallel environment. I think we need to treat that as default, and add ad-hoc fixes to have it run in serial, instead of the other way around. Whenever any pygeo parameterization is used in parallel, we will need that allreduce at least for the total derivatives.

pygeo/DVGeometry.py
```diff
@@ -1953,6 +1952,8 @@ def totalSensitivity(self, dIdpt, ptSetName, comm=None, config=None):
                 dIdx_local[i, :] = self.JT[ptSetName].dot(dIdpt[i, :, :].flatten())
 
         if comm:  # If we have a comm, globaly reduce with sum
+            from mpi4py import MPI
```
anilyil, line 1955: do we have to import mpi4py here?

tests/reg_tests/test_DVGeometryVSP.py
```diff
@@ -13,7 +13,7 @@
     missing_openvsp = True
 
 if not missing_openvsp:
-    from pygeo import DVGeometryVSP
+    from pygeo.DVGeometryVSP import DVGeometryVSP
```
anilyil, line 16: I dont agree with this change and we should import dvgeovsp like others.

pygeo/DVConstraints.py
```diff
@@ -5828,8 +5830,9 @@ def evalCurvArea(self, iSurf):
             # Now compute the KS function for mean curvature, equivelent to KS(H*H*dS)
             sigmaH = np.dot(one, np.exp(self.KSCoeff * H * H * dS))
             KSmean = np.log(sigmaH) / self.KSCoeff
-            if MPI.COMM_WORLD.rank == 0:
-                print("Max curvature: ", max(H * H * dS))
+            # from mpi4py import MPI
```
anilyil, line 5833: Dont leave stuff commented out like this if the code will be removed. Either include if checks so it doesnt run in certain cases, remove the code alltogether, or leave it in there.

anilyil (changes requested):

nwu63:
> I dont think removing mpi4py is necessary like Neil said. We can use a dummy mpi4py like openmdao if its not available. I dont like the separate import of DVGeoVSP. I also dont like that we import MPI during that eval functions sense call.
> 
> This code is designed to work in a parallel environment. I think we need to treat that as default, and add ad-hoc fixes to have it run in serial, instead of the other way around. Whenever any pygeo parameterization is used in parallel, we will need that allreduce at least for the total derivatives.

It's not about MPI being unavailable, but that parts of pyGeo doesn't need MPI and you should not import MPI just because it's there -- see #85 for an example where you'd want to import something from `geo_utils` but because something included in the `__init__` imports MPI, suddenly your script is importing MPI which fundamentally changes certain behaviour.

I take back what I said about the environment variable, that doesn't really make sense IMO. I think the user should be able to import `geo_utils` without having something within pygeo automatically importing MPI, but I'm not sure what's the cleanest way to do that.

pygeo/DVConstraints.py
marcomangano: yeah nevermind,I was thinking about a direct import but that would be bad practice AND would probably get other MPI issues.

nwu63:
I'm going to close this PR -- we decided that this is probably not the best approach. Further discussions are encouraged in #85.

PR #88, opened 2021-07-09
Added coverage

nwu63:
## Purpose
Added codecov coverage, and also enforced that all tests on the private image should pass.

## Type of change
What types of change is it?

- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

### DIFF
--- a/.github/build_real.sh
+++ b/.github/build_real.sh
@@ -1,3 +1,3 @@
 #!/bin/bash
 set -e
-pip install .[testing]
+pip install -e .[testing]
--- a/.github/build_real.sh
+++ b/.github/build_real.sh
@@ -1,3 +1,3 @@
 #!/bin/bash
 set -e
-pip install -e .[testing]
+pip install .[testing]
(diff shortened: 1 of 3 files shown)


.github/build_real.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/bash
 set -e
-pip install .[testing]
+pip install -e .[testing]
```
marcomangano, line 3: I guess my question for [this PR](https://github.com/mdolab/multipoint/pull/16) comes from here: why is this necessary / best practice?

marcomangano: I have been redirected to the instructions, and I see how this makes sense... but at the same time it looks like it is not necessary for the IDWarp PR, I am getting a bit lost

nwu63: Interesting, perhaps this is not needed..

marcomangano (approved):

PR #93, opened 2021-07-25
File splitting

kanekosh:
## Purpose
Closes #81.

Splitted large files: `geo_utils.py, DVConstraints.py, DVGeometry.py`.
```
pygeo/pygeo
|
|-- geo_utils
|   |-- __init__.py
|   |-- all functions previously in geo_utils.py
|
|-- constraints
|   |-- all classes previously in DVConstraints.py
|
|-- geometry
|   |-- all classes previously in DVGeometry.py 
|
|-- __init__.py
|-- other python files
```

- Replaced Error classes in `DVConstraints.py` and `DVGeometry.py` with the baseclasses Error. Also removed the Warning class in `DVConstraints.py` because it was not used anywhere.

## Type of change
What types of change is it?
_Select the appropriate type(s) that describe this PR_

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (non-backwards-compatible fix or feature)
- [ ] Code style update (formatting, renaming)
- [x] Refactoring (no functional changes, no API changes)
- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Checklist
_Put an `x` in the boxes that apply._

- [x] I have run `flake8` and `black` to make sure the code adheres to PEP-8 and is consistently formatted
- [x] I have run unit and regression tests which pass locally with my changes
- [ ] I have added new tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation

PR #94, opened 2021-08-09
File splitting with history

### LINKED ISSUE #81: Reorganize the repo
# Description
There are several unnecessarily large files in this repo:
- `DVConstraints.py` has 6468 lines
- `DVGeometry.py` has 4541 lines
- `geo_utils.py` has 4509 lines

We should move some code into new files. For example, we should have separate files for separate constraint and DV classes instead of having them all in the same file. These files can be stored in separate directories.

hajdik:
## Purpose
Closes #81 
Branched from #93, which was closed to preserve more history

- Split large files: `geo_utils.py`, `DVConstraints.py`, `DVGeometry.py`
```
pygeo/pygeo
|
|-- geo_utils
|   |-- __init__.py
|   |-- all functions previously in geo_utils.py
|
|-- constraints
|   |-- all classes previously in DVConstraints.py
|
|-- variables
|   |-- all classes previously in DVGeometry.py 
|
|-- __init__.py
|-- other python files
```
- Replaced the duplicated `Error` class found throughout to the one contained in `baseclasses`
- Removed unused `Warning` class

## Type of change
- [X] Refactoring (no functional changes, no API changes)
- [ ] Documentation update
- [X Maintenance update
- [ ] Other (please describe)

## Checklist


nwu63:
@hajdik confirmed that the full test suite passed in Docker.

nwu63 (approved):

pygeo/geo_utils/topology.py
```diff
@@ -0,0 +1,1485 @@
...
+    nodesFromFace,
+    FaceCmpObject,
+)
+from .orientation import edgeOrientation, faceOrientation
+from .remove_duplicates import unique, uniqueIndex, pointReduce
+from .index_position import indexPosition1D, indexPosition2D, indexPosition3D
+
+# --------------------------------------------------------------
+#                Topology classes
+# --------------------------------------------------------------
```
joanibal, line 20: Should we move this file to the pygeo level? It isn't really a utility since it is just defining base classes that are used throughout.

pygeo/geo_utils/tripan.py
```diff
@@ -0,0 +1,241 @@
+import numpy as np
+import os
+from .bilinear_map import getBiLinearMap
+
+# --------------------------------------------------------------
+#                Tripan-related class
+# --------------------------------------------------------------
```
joanibal, line 7: Should this be move to the same place as pygeo.py since both are used for geometry generation?

joanibal (changes requested):
Looks good, just a few suggestions about some file locations.

pygeo/geo_utils/topology.py
nwu63: Good point, this appears to be just base classes used by other modules. Let's move it one level up then, and also update the imports/init for that.

pygeo/geo_utils/tripan.py
nwu63: I actually think this can be removed, since we deprecated Tripan I don't think anyone's using this code.

hajdik: I'll remove that file.

pygeo/geo_utils/topology.py
hajdik: I'll move that and fix the imports, it looks like one class is used by pyLayout so there will need to be a PR fixing that import too.

nwu63: Good call, I thought I searched our codes but apparently GitHub's search is inferior to grep. I do agree with @joanibal though, `geo_utils.Topology()` doesn't make a lot of sense.

nwu63 (approved):

joanibal (approved):

PR #97, opened 2021-08-23
Move DVGeometryESP.py for consistent structure

hajdik:
## Purpose
Move `DVGeometryESP.py` into `parameterize` and update imports accordingly. This PR breaks backwards compatibility as it changes how `DVGeometryESP` is imported. 

## Type of change
- [X Breaking change (non-backwards-compatible fix or feature)
- [X] Code style update (formatting, renaming)
- [ ] Documentation update
- [X] Maintenance update
- [ ] Other (please describe)

## Testing
`DVGeometryESP` should be able to be imported through `pyGeo` instead of directly importing the file. 

## Checklist
- [X I have run `flake8` and `black` to make sure the code adheres to PEP-8 and is consistently formatted


nwu63 (approved):

marcomangano (commented):
The PR looks good, but I am now wondering if we want to use the same import approach that @nwu63 used [for the baseclasses refactoring](https://github.com/mdolab/baseclasses/pull/64) for pyGeo as well.
The minor version increase is justified by the backward incompatibility for the ESP wrapper.. but what if we dare more and change all the imports? Or use the temporary "dual" import approach.

(also, some renaming of `DVGeometryVSP.py` and `DVGeometryESP.py` might be appropriate?)

marcomangano:
I fixed the bug when `pip install .` (you either have to explicitly list the subfolders in `setup.py` for them to be copied in `site-packages` - or use a `setuptools` function as done here).
I also updated the `__init__` files to mirror the approach used in https://github.com/mdolab/baseclasses/pull/64

I think we might want to take the opportunity to also rename some of the ambiguous files - e.g DVGeometryVSP is used for both the file and the class

nwu63:
Yeah good changes here, I agree with what you're saying. I just updated the `setup.py` file slightly -- blanket call of `find_packages()` will pick up other packages in the current root directory that will have an init file. For example, if `examples/__init__.py` exists, this will install a package called `examples` in your site-packages directory, which is going to be really annoying to debug in the future. My patch fixes this by including only `pygeo*` packages. @marcomangano please try this locally to make sure it is installing the right files.

As for renaming files, what about the following:
- `DVGeometryVSP` -> `DVGeoVSP`
- `DVGeometryESP` -> `DVGeoESP`
- `DVGeometryAxi` -> `DVGeoAxi`

PR #100, opened 2021-09-06
Code cleanup

nwu63:
## Purpose
This PR cleans up some of the code:
- used pyupgrade to update some syntax
- some manual fixes (replaced `type()` etc.)
- flake8 fixes
- docstring updates

## Type of change
What types of change is it?

- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Testing
N/A

## Checklist

### DIFF
--- a/pygeo/constraints/DVCon.py
+++ b/pygeo/constraints/DVCon.py
@@ -19,7 +19,7 @@ from .planarityConstraint import PlanarityConstraint
 from .curvatureConstraint import CurvatureConstraint
 
 
-class DVConstraints(object):
+class DVConstraints:
     """DVConstraints provides a convenient way of defining geometric
     constraints for WINGS. This can be very useful for a constrained
     aerodynamic or aerostructural optimization. Three types of
@@ -374,7 +374,7 @@ class DVConstraints(object):
             points.append(p1[i])
             points.append(p2[i])
             for i in range(len(points)):
-                f.write("%f %f %f\n" % (points[i][0], points[i][1], points[i][2]))
+                f.write(f"{points[i][0]:f} {points[i][1]:f} {points[i][2]:f}\n")
 
         for i in range(len(p0)):
             f.write("%d %d %d\n" % (3 * i + 1, 3 * i + 2, 3 * i + 3))
@@ -428,7 +428,7 @@ class DVConstraints(object):
         surfaceName="default",
         DVGeoName="default",
     ):
-        """
+        r"""
         Add a set of thickness constraints that span a logically a
         two-dimensional region. A little ASCII art can help here
 
@@ -606,7 +606,7 @@ class DVConstraints(object):
         surfaceName="default",
         DVGeoName="default",
     ):
-        """
+        r"""
         Add a set of thickness constraints oriented along a poly-line.
 
         See below for a schematic
@@ -749,7 +749,7 @@ class DVConstraints(object):
         surfaceName="default",
         DVGeoName="default",
     ):
-        """
+        r"""
         Add a set of leading edge radius constraints. The constraint is set up
         similar to the 1D thickness or thickness-to-chord constraints. The user
         provides a polyline near the leading edge and specifies how many
@@ -1176,7 +1176,7 @@ class DVConstraints(object):
         surfaceName="default",
         DVGeoName="default",
     ):
-        """
+        r"""
         Add a set of thickness-to-chord ratio constraints oriented along a poly-line.
 
         See below for a schematic
@@ -1538,7 +1538,7 @@ class DVConstraints(object):
         surfaceName="default",
         DVGeoName="default",
     ):
-        """
+        r"""
         Add a single volume constraint to the wing. The volume
         constraint is defined over a logically two-dimensional region
         as shown below
@@ -2807,7 +2807,7 @@ class DVConstraints(object):
 
         pts = None
 
-        f = open(fileName, "r")
+        f = open(fileName)
         nSurf = np.fromfile(f, "int", count=1, sep=" ")[0]
         sizes = np.fromfile(f, "int", count=3 * nSurf, sep=" ").reshape((nSurf, 3))
         nElem = 0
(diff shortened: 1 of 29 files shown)


tests/reg_tests/test_DVConstraints.py
```diff
@@ -69,7 +69,7 @@ def evalFunctionsSensFD(self, DVGeo, DVCon, fdstep=1e-2):
                     funcsSens[outkey][inkey][:, array_ind] = deriv_temp
                 xDV[inkey][array_ind] = baseVar[array_ind]
         DVGeo.setDesignVars(xDV)
-        DVCon.evalFunctions(dict())
+        DVCon.evalFunctions({})
```
nwu63, line 72: I've no idea why this line is needed..

marcomangano: You mean `{}` instead of `dict()`? Or the fact that we are calling `DVCon`?

pygeo/parameterization/DVGeoAxi.py
```diff
@@ -240,11 +240,11 @@ def addPointSet(self, points, ptName, origConfig=True, **kwargs):
 
         xform = self.axiTransforms[ptName] = _AxiTransform(points, self.center, self.collapse_into, self.complex)
 
-        super(DVGeometryAxi, self).addPointSet(xform.c_pts, ptName, origConfig, **kwargs)
+        super().addPointSet(xform.c_pts, ptName, origConfig, **kwargs)
```
marcomangano, line 243: I am not very familiar with the use of `super()`, why were the previous arguments unnecessary?

pygeo/parameterization/DVGeoESP.py
```diff
@@ -640,7 +640,7 @@ def writeCADFile(self, filename):
         valid_filetypes = ["brep", "bstl", "egads", "egg", "iges", "igs", "sens", "step", "stl", "stp", "tess", "grid"]
         file_extension = filename.split(".")[-1]
         if file_extension.lower() not in valid_filetypes:
-            raise IOError(
+            raise OSError(
```
marcomangano, line 643: Is `IOError` deprecated? Looks like [OSError](https://docs.python.org/3/library/exceptions.html#OSError) is more general

pygeo/constraints/DVCon.py
```diff
@@ -118,7 +118,7 @@ def setSurface(self, surf, name="default", addToDVGeo=False, DVGeoName="default"
         if name in self.surfaces.keys():
             raise KeyError("Surface names must be unique. Repeated surface name: " + str(name))
 
-        self.surfaces[name] = list()
+        self.surfaces[name] = []
```
marcomangano, line 121: is `list()` too old fashioned?

tests/reg_tests/test_DVConstraints.py
marcomangano: Oh now I see, it's a third call to `DVCon` at the end, just before exiting the FD sensitivity method.. Maybe it's just to reset the internal values, same as done by `DVGeo.setDesignVars()` above? I need to check the function

marcomangano (commented):
Looks good, see my comments/questions

tests/reg_tests/test_DVConstraints.py
marcomangano: @nwu63 I suspect that it is needed because every time we evaluate a constraint we [update the coordinates of the pyGeo object](https://github.com/mdolab/pygeo/blob/9eff5c04b999c8e8f7e99c80b6850868a87415c3/pygeo/constraints/thicknessConstraint.py#L51). I think that if we directly use `DVGeo.update()` the tests should work, up to you if you want to play with that.
I might be wrong bc I don't know what error you are getting.

pygeo/parameterization/DVGeoESP.py
nwu63: IOError is indeed deprecated, from the link above:
```
Changed in version 3.3: EnvironmentError, IOError, WindowsError, socket.error, select.error and 
map.error have been merged into OSError, and the constructor may return a subclass.
```

pygeo/parameterization/DVGeoAxi.py
nwu63: It's an unnecessarily verbose approach (which was mandatory in Python 2 I think). If a class only has one parent, `super()` will refer to it, there is no need to use `super(<class>, self)`

pygeo/constraints/DVCon.py
nwu63: It feels kinda unnecessary, like using `x = str()` instead of `x = ""`.

tests/reg_tests/test_DVConstraints.py
nwu63: Yeah I've no idea, but without line 72 one of the tests fail, and if I also take out line 71 then a few more tests fail. I can see cases where 71 is needed (even though it's probably not the best practice), but not sure why 72 is needed at all.

hajdik: I can't tell what that line is doing either, the error I get from taking it out is in `Test 8: Circularity constraint` and looks like it's not meeting the tolerance. Neil is this what you're seeing?
```
AssertionError: 
Not equal to tolerance rtol=1e-07, atol=1e-07
Failed value for: circularity_base
Mismatched elements: 8 / 9 (88.9%)
Max absolute difference: 2.15268112e-05
Max relative difference: 2.15268112e-05
 x: array([0.999989, 0.999978, 0.999987, 1.      , 0.999995, 0.99998 ,
       0.999981, 0.999994, 1.      ])
 y: array([1., 1., 1., 1., 1., 1., 1., 1., 1.])
```

nwu63: Yes exactly.

marcomangano (approved):
After some offline discussion, I think this is ready to be merged.
That constraint test will really need some overhauling, but it is something that should be done at a later stage

hajdik (approved):
Updates all look good, agree with fixing the constraint test at a later date.

PR #101, opened 2021-09-10
pyGeo.py test

hajdik:
## Purpose
A test has been added to start coverage of `pyGeo.py`. This test is based on the `generate_wing.py` script from the MACH-Aero tutorial. That script is run and then the resulting surface data is summed and compared to a reference file. This is summed rather than directly compared to avoid saving and comparing a long file.  

## Type of change
- [ ] Documentation update
- [X] Maintenance update
- [ ] Other (please describe)

## Testing
Tests should run with latest input files folder

## Checklist


marcomangano:
Nice, the new `BaseRegTest` seems to have fixed the mismatches on different images... but I am not very happy about having a 70k lines ref file. We should find a better way to compare the output file, this is pretty brute force

nwu63:
> Nice, the new `BaseRegTest` seems to have fixed the mismatches on different images... but I am not very happy about having a 70k lines ref file. We should find a better way to compare the output file, this is pretty brute force

Good point, we probably want to do some reduction. `par_add_sum` might be a good candidate, i.e. just sum everything and compare that.

nwu63 (approved):
LGTM

marcomangano (approved):
Nice

Issue #57, opened 2021-01-18
Merge the dvgeomulti code and refactor handling of multiple dvgeos across pygeo


anilyil:
# Description of feature

The PR #52 introduces a way to use multiple DVGeometry objects with a single DVConstraints object. So far, only @bbrelje uses these. There is a separate development of DVGeometryMulti that is used to handle intersecting components. I also implemented a similar functionality but in a different way. In that work, DVGeometryMulti class handles multiple component DVGeometries, and adjusts the naming for pyoptsparse etc. 

Ideally we want to have a consistent implementation so both approaches can function together. 

# Potential solution

Around April 2021, we will refactor the pygeo code with @bbrelje to fix this consistency issue. At the same time, I will also merge the outstanding work in my development branch: https://github.com/anilyil/pygeo/tree/dev The main addition will be the DVGeometryMulti code, but there are other minor API changes for other pygeo classes.

sseraj:
We merged DVGeoMulti in #113 but did not refactor the handling of multiple DVGeo objects with DVCon. I updated the issue title to reflect this.

PR #103, opened 2021-10-03
Child FFD constraint tests and fixes

sseraj:
## Purpose

This PR makes several changes related to DVConstraints and child FFDs:

- Parameterized the DVConstraints tests to also run them with child FFDs. The parent FFD is a stationary box around the original FFD. We use the original reference files for the child FFD tests, which verifies that the constraint values and derivatives are identical with a fixed parent FFD.
The only test that is not parameterized is the triangulated surface (GeoGrad) test with multiple DVGeos because it is unclear how to handle multiple DVGeos with child FFDs (related to #57).

- Fixed `addLinearConstraintsShape` and `addMonotonicConstraints` to work with child FFDs. Linear constraints need a `childIdx` argument because they operate directly on the FFD.

- Added a test for the curvature constraint. This was one of only two constraints that were not tested, the other being the gear post constraint (which I doubt anyone still uses).

- Added a test for the LE radius constraint. This was already tested in `test_Cylinder.py`, but a wing provides a more typical test case. In addition, this test is part of the same parameterized class so it is also tested with child FFDs.

- Renamed the DVConstraints tests and reference files to have actual names instead of numbers.

## Type of change

- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Checklist


tests/reg_tests/test_DVConstraints.py
```diff
@@ -163,7 +186,11 @@ def c172_test_twist(self, DVGeo, DVCon, handler):
         # change the DVs
         xDV = DVGeo.getValues()
         xDV["twist"] = np.linspace(0, 10, self.nTwist)
-        DVGeo.setDesignVars(xDV)
+        if self.child:
+            # Twist needs to be set on the parent FFD to get accurate derivatives
```
marcomangano, line 245: I can confirm this is the same approach I use, but is this documented anywhere? What is the reason behind this?

tests/reg_tests/test_DVConstraints.py
```diff
@@ -810,6 +812,137 @@ def test_14(self, train=False, refDeriv=False):
...
+        xDV = DVGeo.getValues()
+        np.random.seed(37)
+        xDV["local"] = np.random.normal(0.0, 0.05, 32)
+        DVGeo.setDesignVars(xDV)
+        DVCon.evalFunctions(funcs, includeLinear=True)
+        DVCon.evalFunctionsSens(funcsSens, includeLinear=True)
+        handler.root_add_dict("funcs_deformed", funcs, rtol=1e-6, atol=1e-6)
+        handler.root_add_dict("derivs_deformed", funcsSens, rtol=1e-6, atol=1e-6)
+        return funcs, funcsSens
+
```
marcomangano: Is there a way to directly skip the class initialization down here? no biggie though

pygeo/constraints/DVCon.py
```diff
@@ -2007,6 +2017,11 @@ def addLinearConstraintsShape(
 
         self._checkDVGeo(DVGeoName)
 
+        if childIdx is not None:
+            DVGeo = self.DVGeometries[DVGeoName].children[childIdx]
```
marcomangano, line 2021: It took me a bit of code reading, but I see how this makes sense. We have a single DVcon object which "receives" the parent DVGeo, and we were previously applying the constraint to this object instead of the children (when present) as expected. This means that we now need to add one of these for each child ffd, right? Can you add a short description of what is going on and how to extract the `childIdx` index from DVGeo?

tests/reg_tests/test_DVConstraints.py
```diff
@@ -113,6 +113,10 @@ def setUp(self):
         self.base_path = os.path.dirname(os.path.abspath(__file__))
 
     def generate_dvgeo_dvcon_rect(self, addToDVGeo=False):
+        """
+        The rectangular box is primarily used to test unscaled constraint function
+        values against known values for thickness, volume, and surface area.
+        """
```
marcomangano, line 119: Nice! Are the comments on the rest of the functions just redundant, and so you opted to remove then? I think we could use a one-liner for all the function here, even if the names are pretty self explanatory

tests/reg_tests/test_DVConstraints.py
```diff
@@ -732,6 +778,34 @@ def test_monotonic(self, train=False, refDeriv=False):
                 atol=1e-7,
             )
 
+    @unittest.skipIf(missing_geograd, "requires geograd")
```
marcomangano, line 781: So, the other test is on its own only because cannot be used under parameterized?

tests/reg_tests/test_DVConstraints.py
```diff
@@ -807,132 +881,8 @@ def setUp(self):
...
-            funcs, funcsSens = generic_test_base(DVGeo, DVCon, handler, fdstep=1e-3)
-            handler.assert_allclose(
-                funcs["DVCon1_trisurf_constraint_0_KS"], 0.34660627481696404, name="KS", rtol=1e-7, atol=1e-7
-            )
-            handler.assert_allclose(funcs["DVCon1_trisurf_constraint_0_perim"], 0.0, name="perim", rtol=1e-7, atol=1e-7)
-            funcs, funcsSens = self.bwb_test_twist(DVGeo, DVCon, handler)
-            funcs, funcsSens = self.bwb_test_deformed(DVGeo, DVCon, handler)
-
-    @unittest.skipIf(missing_geograd, "requires geograd")
-    def test_triangulatedSurface_intersected(self, train=False, refDeriv=False):
```
marcomangano: Where did this end up?

tests/reg_tests/test_DVConstraints.py
```diff
@@ -870,6 +870,57 @@ def twist(val, geo):
...
+            DVCon.addCurvatureConstraint(surfFile, curvatureType="mean", scaled=False, name="unscaled_curvature_con")
+
+            funcs, funcsSens = generic_test_base(DVGeo, DVCon, handler, fdstep=1e-5)
+            funcs, funcsSens = self.wing_test_twist(DVGeo, DVCon, handler)
+            funcs, funcsSens = self.wing_test_deformed(DVGeo, DVCon, handler)
+
+    def test_LERadius(self, train=False, refDeriv=False):
+        refFile = os.path.join(self.base_path, "ref/test_DVConstraints_LERadius.ref")
+        with BaseRegTest(refFile, train=train) as handler:
+            DVGeo, DVCon = self.generate_dvgeo_dvcon_c172()
```
marcomangano, line 912: Maybe we can generalize the name of this function as well?

marcomangano (commented):
Nice stuff, this test script was really messed up.
Take a look at my comments, I will re-check the final script tomorrow because it is hard to follow these big commits that move function blocks around

tests/reg_tests/test_DVConstraints.py
marcomangano: Despite being outdated, I still refer to the `RegTestGeograd` class

tests/reg_tests/test_DVConstraints.py
sseraj: I don't think it's documented anywhere mostly because manually setting design variables is not too common. This is not specific to twist actually. My understanding is that `setDesignVars` should always be called on the parent, which will recursively set the DVs on all children. It just so happens that shape variables can be applied directly on a child FFD without causing issues. I can change the approach for all DVs to be consistent if you think that's clearer.

tests/reg_tests/test_DVConstraints.py
sseraj: You mean skip the class instead of each test if geograd is not installed? Yes, I think that's possible. I will change that.

pygeo/constraints/DVCon.py
sseraj: Yes, without `childIdx` the constraint is applied on the parent. You only need to add this constraint to each FFD that needs it. `childIdx` is defined by the order in which you add the child FFD to the parent (the first is 0, second is 1, etc.). I will update the docstrings to include this description.

To be clear though, I took this approach directly from [addLeTeConstraints](https://github.com/mdolab/pygeo/blob/7cf8e4e0b827c12cb2f2e79bd28b44bbe9228f15/pygeo/constraints/DVCon.py#L1857-L1860) which already worked with child FFDs.

tests/reg_tests/test_DVConstraints.py
sseraj: Yes, I felt the comments were redundant. Some were also incorrect (saying that a constraint does not rely of point sets when it does). What did you have in mind for the one-liners? I think they would be almost exactly the same as the test name.

tests/reg_tests/test_DVConstraints.py
sseraj: Yes, that's right

tests/reg_tests/test_DVConstraints.py
sseraj: That is [here](https://github.com/mdolab/pygeo/blob/ab236b3eb91586d6fe23278010c13cf0b220a7ec/tests/reg_tests/test_DVConstraints.py#L799). I initially moved all the geograd tests into a separate class but realized I only needed to move the one with 2 DVGeos out.

tests/reg_tests/test_DVConstraints.py
sseraj: Just the name of the functionality as well? I thought about creating one function that takes works with all the geometries because this would reduce code duplication. I can take a look at that if think it's worth it.

sseraj:
Thanks @marcomangano for the thorough review! I will get started on the changes. Feel free to add other suggestions as well

tests/reg_tests/test_DVConstraints.py
marcomangano: I don't think we do it in other repos, but if you find a way it could be nice. The initialization time for pyGeo is not that bad, but the solvers might benefit from this?

tests/reg_tests/test_DVConstraints.py
marcomangano: yeah I see, don't worry, this is fine

tests/reg_tests/test_DVConstraints.py
marcomangano: uh, that would be nice if low-ish effort

marcomangano: Ok, I just checked the functions and I think we should definitely merge `generate_dvgeo_dvcon_c172` and `generate_dvgeo_dvcon_rect`. The only differences are the meshes, the location of the ref axis, the `addToDVGeo`(?) option in `DVCon.setSurface`, and the normalization of the `testMesh.vectors`(??) and they should be pretty easy to handle with an option

tests/reg_tests/test_DVConstraints.py
```diff
@@ -922,6 +831,7 @@ def test_LERadius(self, train=False, refDeriv=False):
             funcs, funcsSens = self.wing_test_deformed(DVGeo, DVCon, handler)
 
 
+@unittest.skipIf(missing_geograd, "requires geograd")
```
marcomangano, line 834: that was easy lol

marcomangano (approved):

ArshSaja (approved):
Nice work!

PR #104, opened 2021-10-12
Several changes to demoDesignVars

sseraj:
## Purpose

I made several changes to demoDesignVars, mainly:

- Added an option to write out the deformed surface mesh by passing in an ADflow object
- Made having a point set optional. If you only want the deformed FFD, there is no need for DVGeo to have a point set.
- Fixed the initialization of the `upper` variable. The function now works if a parent FFD has no DVs.
- Changed the file naming to reset the counter for every DV instead of having one counter for all DVs

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing

In your optimization script, before running the optimizer, put something like the following lines:
```
CFDSolver.setAeroProblem(ap)
DVGeo.demoDesignVars(outDir, CFDSolver=CFDSolver)
```

## Checklist

### DIFF
--- a/pygeo/parameterization/DVGeo.py
+++ b/pygeo/parameterization/DVGeo.py
@@ -2660,8 +2660,9 @@ class DVGeometry:
         callBack : function
             This allows the user to perform an additional task at each new design
             variable iteration. The callback function must take two inputs:
-            1) the output directory name (str) and
-            2) the iteration count (int).
+
+            1. the output directory name (str) and
+            2. the iteration count (int).
         freq : int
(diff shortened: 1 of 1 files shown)


sseraj:
> We don't want to test this with a CFDSolver instance... maybe we can add that test to the `MACH-Aero` tutorial?

Right, I didn't want to introduce ADflow as a testing dependency. Given that this only tests the files created, I don't see much benefit in testing this in MACH-Aero.

Issue #69, opened 2021-02-28
writeRefAxis is midly broken


bbrelje:
## Description
DVGeometry.writeRefAxes appears to have broken midly due to a change in OrderedDict(). It should be reg tested but currently is not

### Steps to reproduce issue
1. Call writeRefAxes in the normal manner

### Current behavior
Raises the following error:
```
Traceback (most recent call last):
  File "runFFDExample.py", line 50, in <module>
    DVGeo.writeRefAxes('local_refaxes.dat')
  File "/home/ben/packages/pygeo/pygeo/DVGeometry.py", line 2066, in writeRefAxes
    print(self.points.keys()[0])
TypeError: 'odict_keys' object is not subscriptable
```

### Expected behavior
Should not do that

bbrelje:
Just needs to be cast to a list, I'll take care of it in my docs push

PR #107, opened 2021-11-03
writeRefAxes fix for Python 3

sseraj:
## Purpose

`.keys()` in Python 3 does not return a list, so `writeRefAxes` has been broken.

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing

I added a test that just checks if files have been written.

## Checklist

### DIFF
--- a/pygeo/parameterization/DVGeo.py
+++ b/pygeo/parameterization/DVGeo.py
@@ -2481,7 +2481,7 @@ class DVGeometry:
            extension will be added.
         """
         # Name here doesnt matter, just take the first one
-        self.update(self.points.keys()[0], childDelta=False)
+        self.update(list(self.points.keys())[0], childDelta=False)
 
         gFileName = fileName + "_parent.dat"
         if not len(self.axis) == 0:
--- a/tests/reg_tests/test_DVGeometry.py
+++ b/tests/reg_tests/test_DVGeometry.py
@@ -980,6 +980,32 @@ class RegTestPyGeo(unittest.TestCase):
         shutil.rmtree(ffdPath)
         shutil.rmtree(pointSetPath)
 
+    def test_writeRefAxes(self):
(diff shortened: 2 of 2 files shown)


marcomangano (approved):
Good catch

marcomangano:
Does this address https://github.com/mdolab/pygeo/issues/69 ?

sseraj:
> Does this address #69 ?

Yes, thanks for catching that.

pygeo/parameterization/DVGeo.py
```diff
@@ -2481,7 +2481,7 @@ def writeRefAxes(self, fileName):
            extension will be added.
         """
         # Name here doesnt matter, just take the first one
-        self.update(self.points.keys()[0], childDelta=False)
+        self.update(list(self.points.keys())[0], childDelta=False)
```
nwu63, line 2484: The order is not guaranteed in Python <3.7, so to be safe I would suggest sorting the list before accessing `[0]`. That way, we are sure all procs have the same name.

Also not sure if this is a concern here or not, I'm not familiar with this part of the code.

sseraj: I don't think it's concern here because we usually call the write functions on a single proc anyway (the underlying `writeTecplot1D` function from pySpline is serial). There are other places in DVGeo where we cast the keys to a list where it might be more important.

In any case, we only support and test Python >= 3.7 right?

nwu63: Yeah fair, if this is expected to be run on a single proc then we are okay. Like you said, this issue might exist elsewhere but that's a separate problem.

Yes, but at the same time I hesitate to rely on behaviour which is built into the underlying CPython implementation rather than something which is built into the language, if that makes sense. In the future they could change this behaviour again and I'd rather be safe than sorry.

nwu63 (approved):

PR #108, opened 2021-12-26
A lot of minor/bug fixes

nwu63:
## Purpose
Basically, I got bored and decided to try to see what pylint would complain about. Turns out, there are loooots of problems with the code that flake8 didn't spot. This is kind of the motivation behind mdolab/.github#22.

Changes include
- simplified many of the init functions used by the different constraints
- made `baseConstraint` into an abstract base class
- removed unnecessary empty `return`
- removed unused variables
- fixed string concatenation
- fixed loop variable used outside scope
- added undeclared class attributes to init
- fixed utterly broken code such as `if np.asscalar:` which should be `if np.asscalar(value):`
- fixed many many typos, including class attribute names
- fixed missing, or extra, or incorrectly arrange function arguments
- simplified comparison logic by chaining comparison operators

I apologize in advance for whoever gets assigned to review this.

## Type of change

- [x] Code style update (formatting, renaming)
- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/pygeo/constraints/DVCon.py
+++ b/pygeo/constraints/DVCon.py
@@ -1378,10 +1378,6 @@ class DVConstraints:
             addToPyOpt=False, the lower, upper and scale variables are
             meaningless
         """
-        try:
-            import geograd  # noqa
-        except ImportError as e:
-            raise ImportError("Geograd package must be installed to use triangulated surface constraint") from e
         if DVGeo_1_name is not None:
             self._checkDVGeo(DVGeo_1_name)
             DVGeo1 = self.DVGeometries[DVGeo_1_name]
--- a/pygeo/constraints/DVCon.py
+++ b/pygeo/constraints/DVCon.py
@@ -369,12 +369,9 @@ class DVConstraints:
         f.write("Nodes = %d, Elements = %d ZONETYPE=FETRIANGLE\n" % (len(p0) * 3, len(p0)))
         f.write("DATAPACKING=POINT\n")
         for i in range(len(p0)):
-            points = []
-            points.append(p0[i])
-            points.append(p1[i])
-            points.append(p2[i])
-            for i in range(len(points)):
-                f.write(f"{points[i][0]:f} {points[i][1]:f} {points[i][2]:f}\n")
+            points = [p0[i], p1[i], p2[i]]
+            for j in range(len(points)):
+                f.write(f"{points[j][0]:f} {points[j][1]:f} {points[j][2]:f}\n")
 
         for i in range(len(p0)):
             f.write("%d %d %d\n" % (3 * i + 1, 3 * i + 2, 3 * i + 3))
@@ -396,8 +393,8 @@ class DVConstraints:
         """
         try:
             from stl import mesh
-        except ImportError:
-            raise ImportError("numpy-stl package must be installed")
+        except ImportError as e:
+            raise ImportError("numpy-stl package must be installed") from e
         if fromDVGeo is None:
             p0, p1, p2 = self._getSurfaceVertices(surfaceName=surfaceName)
         else:
@@ -1383,8 +1380,8 @@ class DVConstraints:
         """
         try:
             import geograd  # noqa
-        except ImportError:
-            raise ImportError("Geograd package must be installed to use triangulated surface constraint")
+        except ImportError as e:
+            raise ImportError("Geograd package must be installed to use triangulated surface constraint") from e
         if DVGeo_1_name is not None:
             self._checkDVGeo(DVGeo_1_name)
             DVGeo1 = self.DVGeometries[DVGeo_1_name]
@@ -1748,12 +1745,10 @@ class DVConstraints:
         for vol in vols:
             try:
                 volCons.append(self.constraints[typeName][vol])
-            except KeyError:
+            except KeyError as e:
                 raise Error(
-                    "The supplied volume name '%s' has not"
-                    " already been added with a call to "
-                    "addVolumeConstraint()" % vol
-                )
+                    f"The supplied volume '{vol}' has not already been added with a call to addVolumeConstraint()"
+                ) from e
         self.constraints[typeName][conName] = CompositeVolumeConstraint(
             conName, volCons, lower, upper, scaled, scale, self.DVGeometries[DVGeoName], addToPyOpt
         )
@@ -2933,7 +2928,7 @@ class DVConstraints:
         # Project all the points
         for i in range(nPts):
             # Project actual node:
-            up, down, fail = geo_utils.projectNode(X[i], axis, p0, p1 - p0, p2 - p0)
+            up, _, fail = geo_utils.projectNode(X[i], axis, p0, p1 - p0, p2 - p0)
             if fail > 0:
                 raise Error(
                     "There was an error projecting a node "
@@ -3096,7 +3091,7 @@ class DVConstraints:
         scalar will be 'upcast' to that size
         """
 
-        if np.isscalar:
+        if np.isscalar(value):
             return value * np.ones((dim1, dim2))
         else:
             temp = np.atleast_2d(value)
@@ -3112,7 +3107,7 @@ class DVConstraints:
         otherwise, a scalar will be 'upcast' to that size
         """
 
-        if np.isscalar:
+        if np.isscalar(value):
             return value * np.ones(dim1)
         else:
             temp = np.atleast_1d(value)
--- a/pygeo/constraints/DVCon.py
+++ b/pygeo/constraints/DVCon.py
@@ -58,7 +58,7 @@ class DVConstraints:
     Parameters
     ----------
     name: str
-        A name for this object. Used to distiguish between DVCon objects
+        A name for this object. Used to distinguish between DVCon objects
         if multiple DVConstraint objects are used in an optimization.
 
     """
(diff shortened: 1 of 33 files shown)


pygeo/constraints/DVCon.py
```diff
@@ -369,12 +369,9 @@ def writeSurfaceTecplot(self, fileName, surfaceName="default"):
         f.write("Nodes = %d, Elements = %d ZONETYPE=FETRIANGLE\n" % (len(p0) * 3, len(p0)))
         f.write("DATAPACKING=POINT\n")
         for i in range(len(p0)):
-            points = []
-            points.append(p0[i])
-            points.append(p1[i])
-            points.append(p2[i])
-            for i in range(len(points)):
-                f.write(f"{points[i][0]:f} {points[i][1]:f} {points[i][2]:f}\n")
+            points = [p0[i], p1[i], p2[i]]
```
marcomangano, line 373: wow this looked bad

pygeo/constraints/DVCon.py
```diff
@@ -396,8 +393,8 @@ def writeSurfaceSTL(self, fileName, surfaceName="default", fromDVGeo=None):
         """
         try:
             from stl import mesh
-        except ImportError:
-            raise ImportError("numpy-stl package must be installed")
+        except ImportError as e:
```
marcomangano, line 397: can you clarify on this fix?

pygeo/constraints/areaConstraint.py
```diff
@@ -68,6 +68,8 @@ def __init__(
         self.perim_scale = perim_scale
         self.max_perim = max_perim
         self.smSize = None
+        self.perim_length = None
```
marcomangano, line 79: Is this to avoid attribute injection at a later stage?

pygeo/constraints/colinearityConstraint.py
```diff
@@ -143,7 +143,7 @@ def addVariablesPyOpt(self, optProb):
         Add the axis variable for the colinearity constraint to pyOpt
         """
 
-        if self.addVarToPyOpt:
+        if self.addToPyOpt:
```
marcomangano, line 146: wow, this also bad

pygeo/constraints/areaConstraint.py
```diff
@@ -290,6 +290,9 @@ def addConstraintsPyOpt(self, optProb):
                 wrt=self.getVarNames(),
             )
 
+    def writeTecplot(self, handle):
```
marcomangano, line 297: Is this function overwritten by children classes?

pygeo/constraints/baseConstraint.py
```diff
@@ -23,13 +24,15 @@ def __init__(self, name, nCon, lower, upper, scale, DVGeo, addToPyOpt):
         self.DVGeo = DVGeo
         self.addToPyOpt = addToPyOpt
 
-    def setDesignVars(self, x):
-        """
-        take in the design var vector from pyopt and set the variables for this constraint
-        This function is constraint specific, so the baseclass doesn't implement anything.
-        """
-        pass
+    # @abstractmethod
```
marcomangano, line 27: I assume this is not use because we don't _always_ need this method? I don't get which criterion you used for abstract methods below, I am not very familiar with abstract classes.

pygeo/geo_utils/dcel.py
```diff
@@ -266,8 +266,8 @@ def buildDcel(self):
 
         for e in self.el:
 
-            h1 = DCELHedge(self.vertices[e.con[0]], self.vertices[e.con[1]], e.X, e.PID, e.uv, e.tag)
-            h2 = DCELHedge(self.vertices[e.con[1]], self.vertices[e.con[0]], e.X, e.PID, e.uv, e.tag)
+            h1 = DCELHedge(self.vertices[e.con[0]], self.vertices[e.con[1]], e.PID, e.uv, e.tag)
```
marcomangano, line 269: This is a straight bug, right? Unused part of the code?

pygeo/parameterization/DVGeo.py
```diff
@@ -1994,7 +1987,7 @@ def totalSensitivity(self, dIdpt, ptSetName, comm=None, config=None):
 
         return dIdx
 
-    def totalSensitivityProd(self, vec, ptSetName, comm=None, child=False, nDVStore=0, config=None):
+    def totalSensitivityProd(self, vec, ptSetName, config=None):
         """
         This function computes sensitivity information.
```
marcomangano, line 1993: Shall we also update the notes at the end of the docstring?

marcomangano (commented):
I gave a pretty superficial check but this looks good, you convinced me about the need for https://github.com/mdolab/.github/pull/22 .
See my comments below

pygeo/constraints/DVCon.py
nwu63: So, this is not necessarily a bug, but more of a convenience thing. Usually when we do
```
except someException:
    raise newException(msg)
```
Python will report two exceptions, which is what happens in the code. It will say something like "during the handling of the above exception, <newException> occured". To avoid this, and make Python raise only a single exception, e.g. `newException`, we must instead do
```
except someException as e:
    raise newException(msg) from e
```
then Python understands that the fact that we got `someException` should be masked and not reported.

pygeo/constraints/areaConstraint.py
nwu63: This attribute is injected somewhere (not in the init), so I've defined it here to make class attributes more explicit.

pygeo/constraints/areaConstraint.py
nwu63: No it is not. This is here to make Python happy because this is an abstract method, meaning it must be implemented by child classes. In this case, I just make an empty implementation because there was no code previously, but in the future this should be implemented.

pygeo/constraints/baseConstraint.py
nwu63: I don't actually fully understand what this is doing, I think it's for constraints that are analytically based on some DVs or something. In any case, this is not needed for all, so I have opted to not include it in the abstract class. Maybe I should just take this out.

pygeo/geo_utils/dcel.py
nwu63: Yeah this is just function signature mismatch, clearly untested code.

pygeo/constraints/DVCon.py
marcomangano: oh that is pretty nice, will try it myself

pygeo/constraints/baseConstraint.py
marcomangano: Yeah I think it's fine to just remove it instead of leave it commented

nwu63:
Pylint works! Results are the same as the local output.

pygeo/parameterization/DVGeo.py
```diff
@@ -16,7 +17,7 @@
 
 
 class DVGeometry:
-    """
+    r"""
```
joanibal, line 20: Why does the string need to be raw?
Is it related to how it is read by the docs generator?

pygeo/constraints/areaConstraint.py
joanibal: Can we raise a not implemented error here?
It's confusing that the function exists but does nothing.

pygeo/constraints/baseConstraint.py
```diff
@@ -55,7 +57,7 @@ def evalFunctionsSens(self, funcsSens, config):
         funcsSens : dict
             Dictionary to place function values
         """
-        pass
+        return
```
joanibal, line 60: I don't understand the reason for the change here. Is it related to the decorator?

joanibal (changes requested):
One small change requested

pygeo/parameterization/DVGeo.py
nwu63: This is sort of a superficial change. By making it raw strings, the LaTeX math stuff no longer need to be escaped with double backslashes when rendered in Sphinx. Either are valid, but this is a bit easier to read in the source code.

pygeo/constraints/areaConstraint.py
nwu63: This is a good point, I will update.

pygeo/constraints/baseConstraint.py
nwu63: Hmm I forget exactly why I made this change---I'll change it back.

nwu63: I remember now: this is because pylint would throw "unnecessary pass statement" warning: https://vald-phoenix.github.io/pylint-errors/plerr/errors/basic/W0107.html but I guess the proper way is to use the docstring as the main part of the code, and simply not put `pass`. It feels weirdly unreadable though. In any case this is something we can figure out later.

joanibal: Yeah, not a problem for me. I was just curious.

pygeo/parameterization/DVGeo.py
joanibal: :+1:

joanibal (approved):

marcomangano (approved):
This is good for me. We should address the remaining warnings at a later stage, right? It looks like some of the function signatures might need to change

nwu63:
> This is good for me. We should address the remaining warnings at a later stage, right? It looks like some of the function signatures might need to change

Yes, the rest will take a bit more time to fix I think, and I felt this PR already had a pretty large diff. I can document some of the more obviously broken things and open some issues.

Issue #110, opened 2022-01-26
Design variables not correctly reset after running `demoDesignVars`


A-CGray:
## Description

After running `demoDesignVars`, the pyGeo design variables appear to be left in whatever state they were in the last written frame. This is an issue because if you then run ADflow, you will be not be the running the baseline geometry you think you are.

### Steps to reproduce issue

1. Setup an ADflow/pyaerostructure case with a DVGeo object
2. Run `demoDesignVars`
3. Run ADflow/pyaerostructure and look at the result in tecplot (the result shown below is from the private MACH tutorial)

### Current behavior
Geometric DV's remain in their last perturbed state.

![Screenshot_20220126_183702](https://user-images.githubusercontent.com/20371123/151265810-2f8d6ed6-a90e-42ef-b7a1-2b787f78ebbd.png)

### Expected behavior
Geometric DV's should be reset to their original values at the end of `demoDesignVars`

I don't understand how this is happening as the DV's appear to be correctly reset in all the intermediate frames, it's only the final perturbed state that is not.

PR #111, opened 2022-01-26
Fix reset issue in `demoDesignVars`

A-CGray:
## Purpose

Closes  https://github.com/mdolab/pygeo/issues/110, I simply save the original DV values in `DVDictOrig` and use this to reset the DV's at the very end of the function

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing

Verified this fixes the problem on case used to demonstrate issue in https://github.com/mdolab/pygeo/issues/110 

## Checklist

### DIFF
--- a/pygeo/parameterization/DVGeo.py
+++ b/pygeo/parameterization/DVGeo.py
@@ -2708,6 +2708,7 @@ class DVGeometry:
 
         # Get design variables
         dvDict = self.getValues()
+        dvDictOrig = self.getValues()
 
         # Loop through design variables on self and children
         geoList = self.getFlattenedChildren()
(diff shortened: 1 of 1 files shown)


pygeo/parameterization/DVGeo.py
```diff
@@ -2788,6 +2789,7 @@ def demoDesignVars(
 
                         # Iterate counter
                         count += 1
+        self.setDesignVars(dvDictOrig)
```
sseraj, line 2792: As you said in #110, the DVs are reset, so you can probably just replace this with `self.setDesignVars(dvDict)` and not have two copies of the dictionary.

A-CGray: You're right, done now

sseraj (approved):

eirikurj (approved):

hajdik (approved):

PR #112, opened 2022-01-28
addition of optional max FFD spline order and optional exclude wrt parameter

anilyil:
## Purpose

This PR introduces two small changes:
1. When creating a DVGeo object using FFDs, we can now specify the maximum order of the splines used in the FFD formulation with the `kmax` parameter, which was hard-coded to 4 previously. A lower order is automatically selected based on the dimensions of the FFD. That behavior is unchanged.
2. Added an optional argument `exclude_wrt` to `addConstraintsPyOpt` in DVCon. This is useful if we want to omit some DVs from geometric constraints manually. 

I am not testing these yet, but I can add tests based on what people suggest. We can also modify existing tests to check for these.

## Type of change
- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist


marcomangano:
The additions look good, thanks for fixing this. Two comments:
- The test failures are happening because of the array order is not the same on different images. We have seen this before but I don't remember if we concluded anything about it. Maybe this is a good opportunity to take care of it?
- I would add just a couple of tests, one in `test_DVconstraints` for the `except_wrt` option and one in `test_DVGeometry` for the spline ordering. I would take one of the existing cases and replicate it with the new options (we should switch to `parameterize` at some point...). For the first case it should be pretty easy, in the second one we need to find a case where the different spline order is creating a noticeable difference, hopefully we can re-use existing FFD files

pygeo/pyBlock.py
```diff
@@ -60,7 +60,7 @@ def __init__(self, initType, fileName=None, FFD=False, symmPlane=None, **kwargs)
     #                     Initialization Types
     # ----------------------------------------------------------------------
 
-    def _readPlot3D(self, fileName, order="f", FFD=False, symmTol=0.001):
+    def _readPlot3D(self, fileName, order="f", FFD=False, symmTol=0.001, kmax=4):
```
nwu63, line 63: I'm guessing this is one of the `kwargs` that gets passed by the `DVGeometry` constructor call? In that case, please document this option both here, and in the `DVGeometry` init call.

pygeo/constraints/baseConstraint.py
```diff
@@ -58,13 +58,23 @@ def getVarNames(self):
         """
         return self.DVGeo.getVarNames()
 
-    def addConstraintsPyOpt(self, optProb):
+    def addConstraintsPyOpt(self, optProb, exclude_wrt=None):
```
nwu63, line 61: Also need docstrings

nwu63 (changes requested):
I was trying to decide if having `exclude_wrt` or `wrt` would be better, but I think the exclude makes more sense as in general we want to include everything. See comments below in addition to what Marco said

pygeo/pyBlock.py
nwu63: I guess this is actually passed through the `pyBlock` constructor as well. I don't like undocumented `kwargs` so we should try to do a better job here.

joanibal:
I suggest making the order in each direction a parameter. 
If it's alright with you I'll update the PR with this feature, which I have already implemented on my dev fork.

joanibal:
I'm good with the PR after Neil's comments have been addressed.

joanibal:
Anil and I talked over slack about removing the option to have different orders in different directions.

joanibal (dismissed):

pygeo/pyBlock.py
anilyil: tried to clarify how this is done a bit by exposing the kmax in the dvgeo init and the pyblock init. please take a look and let me know if its okay now

pygeo/constraints/baseConstraint.py
anilyil: added. let me know if you want any other changes here

anilyil:
I addressed the comments on docstrings. will work on adding the tests now

anilyil:
I added a test on the custom FFD spline order. It fails when the order is not set correctly. I am not sure how I can test the `exclude_wrt` option. Do we test the `addConstraintsPyOpt` methods? If not testing the `exclude_wrt` stuff is okay, this one is good for another review.

joanibal (approved):

nwu63 (approved):
LGTM

nwu63:
> I added a test on the custom FFD spline order. It fails when the order is not set correctly. I am not sure how I can test the `exclude_wrt` option. Do we test the `addConstraintsPyOpt` methods? If not testing the `exclude_wrt` stuff is okay, this one is good for another review.

I'm okay with not testing the `exclude_wrt` option for the time being.

Issue #89, opened 2021-07-10
Nondeterministic test failure


nwu63:
## Description
Sometimes when you run the tests, `test_DVGeometry.py:RegTestPyGeo.test_21` will randomly fail. The nature of the failure is that some entries in the numpy array in the dictinoary `dIdx: mainAxis: x_axis` are swapped. It's unclear why or how, but maybe it has to do with the way we loop over dictionary keys or something.

### Steps to reproduce issue
Unfortunately not reliably reproducible :( but this happened on some of the CI runs of #88 and was observed previously by @eirikurj.

nwu63:
I am able to now reproduce this failure reliably on my local machine, but the fact that buildbot passes means this is not reproducible on Docker.

anilyil:
My ARM images replicated this issue 3 out of 3 times I tried, so I will look into this. Mentioned this in mdolab/docker#169

nwu63:
This is most curious. I installed `numpy==1.19.2` locally and now the error's gone, but the error's there with `numpy=1.19.5`.

anilyil:
I have `numpy==1.21.5` and `scipy==1.2.1` on my arm image and the pygeo test fails consistently. It's on my list to do.

sseraj:
Can we skip this test until the issue is resolved? Pipelines are now failing regularly.

anilyil:
I am working on a fix for this now. Will disable the test if I cannot figure it out.

PR #113, opened 2022-02-03
DVGeometryMulti

sseraj:
## Purpose

This PR adds DVGeometryMulti, the component-based surface mesh deformation method developed by @anilyil as described in [this paper](http://websites.umich.edu/~mdolaboratory/pdf/Yildirim2021b.pdf), with some modifications by me and @yqliaohk to get it 'production ready'.

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing

The test coverage should be pretty good. test_DVGeometryMulti.py tests the intersection functionality and derivatives. I also expanded the parameterization in test_DVConstraints.py to test all the constraints with a multi component setup.

## Checklist


sseraj:
The failed tests are because of #89

sseraj:
>One question: do we want to also create a PR to parallel solvers (at least adflow) that add point sets in parallel?

I'm planning on opening a PR on ADflow once this is merged. I realized that the changes to the `addPointSet` call is not backwards compatible with older pyGeo in certain cases (specifically with child FFDs), so I wanted to hold off on opening the PR until this is merged.

On a related note, I will bump the minor version of pyGeo in this PR so that any backwards incompatibility with older versions is clear in the next ADflow release.

pygeo/parameterization/DVGeo.py
```diff
@@ -592,7 +592,11 @@ def addPointSet(self, points, ptName, origConfig=True, **kwargs):
             Flag determine if the coordinates are projected into the
             undeformed or deformed configuration. This should almost
             always be True except in circumstances when the user knows
-            exactly what they are doing."""
+            exactly what they are doing.
+        comm : None
+            Comm that is associated with the added point set. Does not
+            work now, just added to be consistent with the API of
+            other DVGeo types."""
```
marcomangano, line 600: So does the point set have a comm? I am not sure about how pyGeo handles parallelization.
Any plans to fix/improve this at some point?

pygeo/pyBlock.py
```diff
@@ -789,13 +789,16 @@ def attachPoints(self, coordinates, ptSetName, interiorOnly=False, faceFreeze=No
             kwargs pass through to the actual projectPoints() function
         """
 
+        # Unpack kwargs
+        nIter = kwargs.get("nIter", 100)
+
```
marcomangano, line 794: Given that there are no additional kwargs, we could just pass `nIter` as explicit argument to `atttachPoints`

marcomangano (commented):
I gave a quick look at the parts of the PR, great effort. I have two additional comments:
- It would be great to have a section of the tutorial with a working example. I get the overall idea but there are a lot of additional features (e.g. intersection management) that will require some explanation
- How is the newly added file doing in terms of code duplication? Is it going to be part of the planned code overhauling or is it already leveraging on other existing mehtods/classes?

sseraj:
Thanks for taking the time to review this one @marcomangano; it's not an easy one. 

> It would be great to have a section of the tutorial with a working example. I get the overall idea but there are a lot of additional features (e.g. intersection management) that will require some explanation

I agree, and this is in the works. Given that this is an advanced feature and relies of pySurf, we opted to put it in the private tutorial. Have a look at the `intersection` branch for the work in progress.

> How is the newly added file doing in terms of code duplication? Is it going to be part of the planned code overhauling or is it already leveraging on other existing mehtods/classes?

DVGeometryMulti mimics the API of DVGeometry for compatibility, but there is not much code duplication. In many cases, DVGeometryMulti functions will loop over components and call the corresponding function on the DVGeometry object. See `getValues` for an example.

pygeo/parameterization/DVGeo.py
sseraj: I will mostly defer this question to @anilyil.

To give some context though, the motivation behind adding this is because we want to add `comm` to the `addPointSet` call in ADflow so that DVGeometryMulti can use it. I think we can remove this though because `addPointSet` already takes in arbitrary kwargs and just ignores the ones it doesn't use.

pygeo/pyBlock.py
sseraj: This is related to the `comm` kwargs discussion above. `attachPoints` is called in `addPointSet`
https://github.com/mdolab/pygeo/blob/70ae801f137fe7c0be1d4a72214842e4c8ab0f29/pygeo/parameterization/DVGeo.py#L616-L619
We can make it can explicit argument in attachPoints, but then we simply move the unpacking up one level to `addPointSet`.

pygeo/parameterization/DVGeo.py
anilyil: The FFD-based DVGeos do not really need a communicator; the only parallel operation is the allreduce after the sensitibity computation, and that method takes in an optional comm. I added the comm to the addpointset call because some of the operations on the pointsets need to be comm-aware. 

There are 2 types of pointsets added to dvgeo: serial (duplicated) or distributed. The "serial" pointsets are identical across all procs and this serial computation is basically duplicated on all procs. When this is done, we dont need a comm, because all procs own the entire copy. These points are added by DVCon, so they would not get the intersection treatment in dvgeo anyways.

The parallel points are what sth like adflow adds. These are distributed among adflow's communicator and each processor gets part (or none) of the points. The FFD-based computation does not need to know the global collection of points, so we don't need a comm except until that final allreduce in the sensitivity computation. 

However, the intersection method does need to know the global collection of nodes. This is because the intersection code will need to be aware of all nodes for a particular step in the algorithm (the curve based projection and warping, sections 2.7 and 2.8 in the paper). Because of this, we are adding this optional argument; the "serial" pointsets dont use it, but the distributed pointsets do use it. On the pyGeo side, only the DVGeoMulti uses it.

Regarding absorbing it into the kwargs; if thats already there then I am fine with it since other dvgeos dont use it and we can leave it explicitly defined in the dvgeomulti code

anilyil:
The changes besides the DVGeoMulti code itself look good to me, but someone other than me should also take a look. Thanks again for all the work here @sseraj @yqliaohk

sseraj:
> The changes besides the DVGeoMulti code itself look good to me, but someone other than me should also take a look. 

I agree, I will leave this review to:
@marcomangano: mainly to review the parts outside of DVGeoMulti.py as you have already done, particularly the tests to see if anything can be improved there 
@yqliaohk: to review the tests and DVGeoMulti.py more closely, particularly the docstrings and code clarity

pygeo/pyBlock.py
marcomangano: I see, this is fine as is then

pygeo/parameterization/DVGeo.py
marcomangano: Thanks for the explanation! I see where this goes, but I am still not sure if we want to add a "dead" attribute to this method. I am also wondering if having functions with the same name and "slightly" different behavior in different files (e.g., it looks like we now have three `addPointSet` in pygeo) can cause problems. I assume that once you define the DVGeo object then there are no ambiguities, I am just not sure if this is good practice

marcomangano: I also noticed that for consistency you should add `compNames` here. Do we want to add a warning in case these two arguments are passed in here by mistake. something like "method called with XXX attribute, but XXX is not implemented on DVGeo so it will be ignored"

tests/reg_tests/test_DVGeometryMulti.py
```diff
@@ -0,0 +1,222 @@
...
+
+        # Now we will test the derivatives
+
+        # Build a dIdpt array
+        # We will have nNodes*3 many functions of interest
+        nNodes = pts.shape[0]
+        dIdpt = np.zeros((nNodes * 3, nNodes, 3))
+
+        # Set the seeds to one in the following fashion
+        # The first function of interest gets the first coordinate of the first point
```
marcomangano, line 176: It took me a bit to figure out what is going on here, we are basically getting "individual" derivative outputs for every node coordinate, right? I am not sure if it is worth adding some clarification, especially on a test script though

marcomangano (dismissed):
Just some additional questions on the test for now, but overall is looking good and I do not have major recommendations

pygeo/parameterization/DVGeo.py
sseraj: I removed `comm` from this function and similarly will not add `compNames`.

For the warning, I think it is rare for the user to call `addPointSet` directly. This is mainly used by DVCon and our solvers. DVCon now passes `compNames` for every point set based constraint, so having a warning every time would be annoying.

As for whether this is good practice, I can't really say. The reason we do it is mainly to reduce code duplication in DVCon. It would be cumbersome to add code to each constraint to check what type of DVGeo was passed in and then use the appropriate `addPointSet` call.

tests/reg_tests/test_DVGeometryMulti.py
```diff
@@ -0,0 +1,222 @@
...
+        )
+
+        # Add the point set
+        ptSetName = "test_set"
+        comm = MPI.COMM_WORLD
+        DVGeo.addPointSet(pts, ptSetName, comm=comm, applyIC=True)
+
+        # Apply twist to the two intersecting boxes
+        dvDict = DVGeo.getValues()
+        dvDict["box1_twist"] = 2
```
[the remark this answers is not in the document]
sseraj: No reason really. The twists are only applied on one end of each box, so the intersection curve is changing. It's not a rigid body rotation if that was the concern.

tests/reg_tests/test_DVGeometryMulti.py
sseraj: Yes, that's correct. I will add a comment to clarify this.

tests/reg_tests/test_DVGeometryMulti.py
```diff
@@ -0,0 +1,222 @@
...
+        # Set up DVGeometryMulti object
+        DVGeo = DVGeometryMulti()
+        DVGeo.addComponent("box1", DVGeoBox1, triMeshFiles[0])
+        DVGeo.addComponent("box2", DVGeoBox2, triMeshFiles[1])
+        DVGeo.addComponent("box3", DVGeoBox3, None)
+
+        # Define some feature curves
+        featureCurves = [
+            # Curves on box1
+            "part_15_1d",
```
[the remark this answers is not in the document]
marcomangano: Ok I see, so this will be covered in the documentation at some point :)

yqliaohk (dismissed):
Looks good to me. I also ran the tests locally and they all passed. Thanks for the work!

pygeo/parameterization/DVGeoMulti.py
```diff
@@ -0,0 +1,3146 @@
...
+from collections import OrderedDict
+import numpy as np
+from mpi4py import MPI
+from baseclasses.utils import Error
+from pysurf import intersectionAPI, curveSearchAPI, utilitiesAPI, adtAPI, tsurf_tools, tecplot_interface
+
+
+class DVGeometryMulti:
+    """
+    A class for manipulating multiple components using multiple FFDs.
```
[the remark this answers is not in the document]
sseraj: Good point. I updated the docstring.

tests/reg_tests/test_DVConstraints.py
```diff
@@ -93,12 +100,22 @@ def generic_test_base(DVGeo, DVCon, handler, checkDerivs=True, fdstep=1e-4):
...
+            "multi": False,
         },
         {
+            # Deforming child FFD with a stationary parent FFD
             "name": "child",
             "child": True,
+            "multi": False,
+        },
+        {
+            # One deforming component FFD and a non-intersecting stationary component FFD
```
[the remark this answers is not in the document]
sseraj: That's correct. Let me know if the updated description is clearer.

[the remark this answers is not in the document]
yqliaohk: I see, the updated description is clearer, but why is a not-intersected case chosen as the test case here?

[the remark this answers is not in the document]
sseraj: The idea behind parameterizing the DVConstraints tests is to use the same reference values for all the tests. This checks that the function values and gradients for a single component of DVGeometryMulti are the same as one instance of DVGeometry.

Also, `applyIC` is False for all constraint point sets. This is because we want constraints like thickness and volume to follow the component geometry, not the surface mesh. So even if the components intersected, there would be no intersection treatment applied to the point sets.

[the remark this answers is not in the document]
yqliaohk: I see. Thanks for the clarification!

yqliaohk (approved):

marcomangano (approved):

PR #114, opened 2022-02-11
fix the argsort that caused inconsistent results due to machine precision errors

anilyil:
## Purpose

This PR addresses #89. The underlying cause was an argsort flipping the order of some entries due to the differences on the order of machine precision. By rounding the array to 8 significant figures before the sorting algorithm, we prevent this behavior. 

This might cause some existing runscripts to get a different ordering but as we saw with the test case that started failing, this means that these runscripts were susceptible to this issue anyways. If it shows up, this change will show up as a shuffled ordering of the FFD control points in shape variables. 

## Type of change
- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
We'll see what azure says

## Checklist


marcomangano (approved):
This makes a lot of sense - and there is only one way to figure out if the issue is fully addressed!

PR #116, opened 2022-02-15
Embedding tolerance update

marcomangano:
## Purpose

The arbitrary tolerance currently set to [determine if a point is embedded when interiorOnly is True](https://github.com/mdolab/pygeo/blob/ed15c848703a90055d38130b6d05cef8080a9d68/pygeo/pyBlock.py#L815) (children FFDs most commonly use this option) was creating issues in some of my cases, leaving some points stationary as the rest of the mesh/geometry object is deformed:

This PR addressed this issue doing two things:
- Updates the tolerance `embTol` to `1e-10` instead of `eps * 50` (with `eps` being the tolerance for the Newton search in the projection subroutine in pySpline) and decouples it from `eps`;
- enables the user to pass `embTol` as an argument for easier tuning of the mesh deformation properties.

Note that, as `attachPoints` is called from [addPointSet](https://github.com/mdolab/pygeo/blob/ed15c848703a90055d38130b6d05cef8080a9d68/pygeo/parameterization/DVGeo.py#L578), which is in turn usually called from other solvers methods (see [how it's done in ADflow](https://github.com/mdolab/adflow/blob/007c8a8ed837bf580c02b0507a46d8704aaa3c24/adflow/pyADflow.py#L2690)), other sibling PRs should be opened to propagate the new option to the user level.

Co-authored with @sseraj 

## Type of change

- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Checklist


sseraj (commented):
This looks fine to me, but I'll leave the approvals to others because I was involved in making the changes

eirikurj (changes requested):
Minor comments, let me know if this is unclear.

eirikurj:
@marcomangano I agree that the appropriate docstrings should be where the function parameters are defined/given, and thats all fine. 
But, I just dont see the need to extract the values just to pass it on explicitly, unless its being used in that function or has some purpose, which it is not in this case. I feel this is not what the user expects, and is providing different defaults if the value is not given through the `kwags`. In that case, I would prefer an explicit function signature instead of relying on `kwargs`. There are however some implications by both approaches.
Instead of writing something long here, I think we could have a quick chat offline to discuss this in case I am missing something. I can also make quick edits and just push to a separate branch to show you what I had in mind.

eirikurj:
@marcomangano I pushed what we discussed offline, here are the two branches that have the changes/suggestions that we talked about. Would be good to have some discussion.

- [Suggestion 1](https://github.com/mdolab/pygeo/compare/ffd-bugfix...eirikurj:ffd-bugfix-patch1) - This propagates the `kwargs` downstream. Benefits are that you dont have to define the downstream parameters like `eps` in other functions, they only appear in the function that they are defined. The downside is that potentially garbage `kwargs` are propagated all downstream, (like `faceFreeze` in the test). 
- [Suggestion 2](https://github.com/mdolab/pygeo/compare/ffd-bugfix...eirikurj:ffd-bugfix-patch2) - This has only `addPointSet` to take in `kwargs` since that is the only one that needs it, but the rest has explicit signatures. The benefits here that everything is explicit, and all defaults defined in the function. The downside is that the defaults are defined in several places.

marcomangano:
> @marcomangano I pushed what we discussed offline, here are the two branches that have the changes/suggestions that we talked about. Would be good to have some discussion.
> 
>     * [Suggestion 1](https://github.com/mdolab/pygeo/compare/ffd-bugfix...eirikurj:ffd-bugfix-patch1) - This propagates the `kwargs` downstream. Benefits are that you dont have to define the downstream parameters like `eps` in other functions, they only appear in the function that they are defined. The downside is that potentially garbage `kwargs` are propagated all downstream, (like `faceFreeze` in the test).
> 
>     * [Suggestion 2](https://github.com/mdolab/pygeo/compare/ffd-bugfix...eirikurj:ffd-bugfix-patch2) - This has only `addPointSet` to take in `kwargs` since that is the only one that needs it, but the rest has explicit signatures. The benefits here that everything is explicit, and all defaults defined in the function. The downside is that the defaults are defined in several places.

Thanks for putting this together!!! From my point of view option 1 is better - and we can add a warning/check on `**kwargs` to be sure the user is aware of the garbage `kwargs`. What do you think about it?

marcomangano:
@eirikurj check the second-to-last commit, if we remove `**kwargs` from `projectpoints()` we should catch any unwanted kwarg in that function

eirikurj:
@marcomangano I dont think you can remove the `**kwargs` from `projectPoints` since `addPointSet` needs to be able to take in whatever to support the DVGeoMulti. The tests are failing due to that.

sseraj:
@eirikurj After some discussion, we decided to go for an approach closer to your Suggestion 2. Following a suggestion from @nwu63, I removed all keyword arguments from `projectPoints` to avoid duplicate definitions. All kwargs are now defined in `attachPoints`. Please check that this makes sense.

@marcomangano This should also catch unused kwargs. Popping `compNames` in `addPointSet` explicitly excludes it.

pygeo/parameterization/DVGeo.py
```diff
@@ -600,6 +613,9 @@ def addPointSet(self, points, ptName, origConfig=True, **kwargs):
 
         """
 
+        # compNames is only needed for DVGeometryMulti, so remove it if passed
+        kwargs.pop("compNames", None)
```
eirikurj, line 617: Just a minor concern. DVGeoMulti has more arguments than just `compNames`, like `applyIC` and `comm`. While no constraints that are calling `addPointSet` are including these parameters at the moment, they may. Should we pop them here? If not, the following call to `attachPoints` will fail. To avoid that, you are basically back to my Suggestion 2. Not sure if this is a concern. @sseraj you may have more insights into if that is possibly needed for the constraints when using DVGeoMulti.

marcomangano: I am fine with popping all the unnecessary `kwargs` here

sseraj: I don't anticipate adding the other kwargs like `comm` and `applyIC` to the constraints. If we do end up adding them in the future, we can also pop them. I would like to keep this list to a minimum though just so we catch any potential errors.

eirikurj (approved):

marcomangano:
@joanibal could you take a look at this?

joanibal:
Sure, sorry I didn't know who was responsible for this review

pygeo/pyBlock.py
```diff
@@ -785,34 +785,29 @@ def attachPoints(self, coordinates, ptSetName, interiorOnly=False, faceFreeze=No
...
 
-        # Unpack kwargs
-        nIter = kwargs.get("nIter", 100)
+        """
 
         # Project Points, if some were actually passed in:
         if coordinates is not None:
             if not interiorOnly:
-                volID, u, v, w, D = self.projectPoints(coordinates, checkErrors=True, eps=eps, nIter=nIter)
+                volID, u, v, w, D = self.projectPoints(coordinates, True, embTol, eps, nIter)
```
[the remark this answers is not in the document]
marcomangano: We had quite a long discussion about removing kwargs there, I would rather add a local var `checkErrors : bool` in this function and pass that instead  of the bare `True` of `False` - would it work?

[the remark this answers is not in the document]
sseraj: I like the idea of adding a `checkErrors` local variable. My proposal would be to do `checkErrors = not interiorOnly` and move the `projectPoints` calls outside the `if not interiorOnly` block.

joanibal (approved):

sseraj:
There are some scenarios where we have empty point sets (like in parallel when some procs have points and others don't). I think this is fine as is because `attachPoints` is only used internally.

sseraj (approved):
@marcomangano I think you can merge this. The last change was fairly minor.

PR #118, opened 2022-02-26
Moved some functions out of class into `geo_utils`

nwu63:
## Purpose
There are a lot of geometric functions that were attached to specific constraint classes. This PR moves them (and a few other miscellaneous functions) out of classes since they don't need to be, and I put them in geo_utils. This way, these functions can be accessed outside of these classes if you need them. As an example, the `volpym` function is duplicated in other repos, but now can be directly imported instead. This removes code duplication and ensures any bug fixes are propagated to all codes.

## Expected time until merged
1 week

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/pygeo/constraints/areaConstraint.py
+++ b/pygeo/constraints/areaConstraint.py
@@ -5,6 +5,7 @@ import numpy as np
 from .. import geo_utils
 from mpi4py import MPI
 from .baseConstraint import GeometricConstraint
+from ..geo_utils.polygon import areaTri
 
 try:
     from geograd import geograd_parallel  # noqa
@@ -327,7 +328,7 @@ class SurfaceAreaConstraint(GeometricConstraint):
         self.DVGeo.addPointSet(self.p2, self.name + "p2", compNames=compNames)
 
         # compute the reference area
-        self.X0 = self._computeArea(self.p0, self.p1, self.p2)
+        self.X0 = areaTri(self.p0, self.p1, self.p2)
 
     def evalFunctions(self, funcs, config):
         """
@@ -343,7 +344,7 @@ class SurfaceAreaConstraint(GeometricConstraint):
         self.p1 = self.DVGeo.update(self.name + "p1", config=config)
         self.p2 = self.DVGeo.update(self.name + "p2", config=config)
 
-        self.X = self._computeArea(self.p0, self.p1, self.p2)
+        self.X = areaTri(self.p0, self.p1, self.p2)
         if self.scaled:
             self.X /= self.X0
         funcs[self.name] = self.X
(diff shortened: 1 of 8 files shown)


pygeo/constraints/areaConstraint.py
```diff
@@ -508,7 +490,7 @@ def evalFunctions(self, funcs, config):
         self.p1 = self.DVGeo.update(self.name + "p1", config=config)
         self.p2 = self.DVGeo.update(self.name + "p2", config=config)
 
-        self.X = self._computeArea(self.p0, self.p1, self.p2, self.axis)
+        self.X = self._computeProjectedAreaTri(self.p0, self.p1, self.p2, self.axis)
```
joanibal, line 493: Is there a change in behavior here? Why is it the projected area now? Was it wrong before?

pygeo/geo_utils/polygon.py
```diff
@@ -1,15 +1,39 @@
...
-    # Return the area of the polygon. Note that the input need not be
-    # strictly a polygon, (closed curve in 2 dimensions.) The approach
-    # we take here is to find the centroid, then sum the area of the
-    # 3d triangles. THIS APPROACH ONLY WORKS FOR CONVEX POLYGONS!
+    """Return the area of the polygon.
+    Note that the input need not be strictly a polygon (closed curve in 2 dimensions).
+    The approach we take here is to find the centroid, then sum the area of the
+    3d triangles.
+
+    .. warning:: This approach only works for convex polygons
```
joanibal, line 35: I'm not sure I understand why this is. It is not necessary, but if you have additional insight could you add it to the warning.

joanibal (commented):
1 question and 1 request

pygeo/constraints/areaConstraint.py
marcomangano: (I think the issue is what I mentioned above, but let's wait for Neil's feedback)

nwu63: I renamed the function here because I felt `_computeArea` is too vague and generic for what it actually does, and also to distinguish with the other functions that previously were called `computeArea`.

pygeo/geo_utils/polygon.py
nwu63: I just converted the previous block comment to docstrings, and changed the all caps warning into a warning block. I'm not sure whether the warning is actually true or not, but if you think it is not the case then I can remove it.

pygeo/constraints/areaConstraint.py
```diff
@@ -421,25 +422,6 @@ def evalFunctionsSens(self, funcsSens, config):
 
             funcsSens[self.name] = tmpTotal
 
-    def _computeArea(self, p0, p1, p2):
```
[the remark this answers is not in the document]
nwu63: That is correct, but they were doing different things and the name was very vague. I didn't move the other one because it relied on data from the class itself (i.e. it was using `self`).

marcomangano (approved):

pygeo/geo_utils/polygon.py
joanibal: I don't know if it true, so i guess it is best to leave it be.

joanibal (approved):

PR #119, opened 2022-03-11
Add composite DVs

nwu63:
## Purpose
This PR adds composite DVs, which are linear combinations of existing DVs.

## Expected time until merged

Somewhat urgent, by end of the week please.

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing

Tests will come at a later time.

## Checklist


pygeo/constraints/thicknessConstraint.py
```diff
@@ -65,8 +66,8 @@ def evalFunctionsSens(self, funcsSens, config):
             for i in range(self.nCon):
                 p1b, p2b = geo_utils.eDist_b(self.coords[2 * i, :], self.coords[2 * i + 1, :])
                 if self.scaled:
-                    p1b /= self.D0[i]
-                    p2b /= self.D0[i]
+                    p1b /= self.D0[i].real
+                    p2b /= self.D0[i].real
```
marcomangano, line 70: Is this a bug?

marcomangano: `self.D0` is the [initial value of the thickness](https://github.com/mdolab/pygeo/blob/f0595685c539a946b3b8c5c64364673d1dc8aa65/pygeo/constraints/thicknessConstraint.py#L31) at the evaluation points.. in which case is this going to have an imaginary component? can the distance be negative?

pygeo/parameterization/DVGeo.py
```diff
@@ -1793,7 +1825,7 @@ def pointSetUpToDate(self, ptSetName):
         else:
             return True
 
-    def convertSensitivityToDict(self, dIdx, out1D=False):
+    def convertSensitivityToDict(self, dIdx, out1D=False, compName=False):
         """
```
marcomangano, line 1829: Please update docstring

pygeo/constraints/thicknessConstraint.py
nwu63: Hmm it's possible that `self.D0` can stay as real, I don't fully remember why I made this specific change, but the changes in this file are for fixing CS. Maybe the main thing to be fixed is to use the CS-safe version of L2 norm when computing the perturbed distances.

In any case, adding `self.dtype` seems to have broken some tests, so I'll clean up this stuff soon.

pygeo/parameterization/DVGeo.py
```diff
@@ -1858,14 +1890,22 @@ def convertSensitivityToDict(self, dIdx, out1D=False):
...
                 else:
                     dIdxDict[key] = childdIdx[key]
 
+        # replace other names with user
+        if compName and self.useComposite:
+            array = []
+            for key, val in dIdxDict.items():
+                array.append(val)
+            array = np.hstack(array)
+            dIdxDict = {self.DVComposite.name: array}
```
marcomangano, line 1907: If I get this right, here you are just renaming the key of the dictionary, while the actual mapping happens in `mapSensToComp`? basically when using a composite variable we are still populating the dictionary as we have regular DVs, but the actual entries are not related to the specific DV anymore after the mapping.

pygeo/parameterization/DVGeo.py
```diff
@@ -1858,14 +1890,22 @@ def convertSensitivityToDict(self, dIdx, out1D=False):
...
+            childdIdx = self.children[iChild].convertSensitivityToDict(dIdx, out1D=out1D, compName=compName)
             # update the total sensitivities with the derivatives from the child
             for key in childdIdx:
                 if key in dIdxDict.keys():
                     dIdxDict[key] += childdIdx[key]
                 else:
                     dIdxDict[key] = childdIdx[key]
 
+        # replace other names with user
+        if compName and self.useComposite:
```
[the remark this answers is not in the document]
marcomangano: Related: Is there a situation where `self.useComposite` is True, but `compName` is None?

pygeo/parameterization/DVGeo.py
```diff
@@ -2446,6 +2498,38 @@ def addVariablesPyOpt(
...
         )
+
+        # we add the composite DVs, and construct linear constraints that replace the existing bounds
+        # then we simply return without adding any of the other DVs
+        if self.useComposite:
+            dv = self.DVComposite
+            optProb.addVarGroup(dv.name, dv.nVal, "c", value=dv.value, lower=dv.lower, upper=dv.upper, scale=dv.scale)
+
+            # add the linear DV constraints that replace the existing bounds!
+            # Note that we assume all DVs are added here, i.e. no ignoreVars or any of the vars = False
```
marcomangano, line 2509: Is this a limitation? I can't picture a production case where we use `ignoreVars`, but maybe we can add a warning message to stay on the safe side?

pygeo/parameterization/DVGeo.py
```diff
@@ -2446,6 +2498,38 @@ def addVariablesPyOpt(
...
+
+            optProb.addConGroup(
+                f"{self.DVComposite.name}_con",
+                self.getNDV(),
+                lower=lb,
+                upper=ub,
+                scale=1.0,
+                linear=True,
+                wrt=self.DVComposite.name,
+                jac={self.DVComposite.name: self.DVComposite.u},
```
marcomangano, line 2529: I trust your math here, see comment above

pygeo/parameterization/DVGeo.py
```diff
@@ -2446,6 +2498,38 @@ def addVariablesPyOpt(
...
+            lb = {}
+            ub = {}
+            for lst in varLists:
+                for key in varLists[lst]:
+                    dv = varLists[lst][key]
+                    lb[key] = dv.lower
+                    ub[key] = dv.upper
+
+            lb = self.convertDictToSensitivity(lb)
+            ub = self.convertDictToSensitivity(ub)
```
marcomangano, line 2519: I am not sure I understand the trick here, can you elaborate real quick? you just turn the "regular DV" bounds into upper and lower constraint values, and use the singular vector matrix as Jacobian bc this is a linear transformation?
Maybe I am just confused about how pyOptSparse handles linear constraints, but it looks like we are not _evaluating_ them here, right?

pygeo/parameterization/DVGeo.py
```diff
@@ -3363,6 +3447,39 @@ def _unComplexifyCoef(self):
 
             self.coef = self.coef.real.astype("d")
 
+    def mapXDictToDVGeo(self, inDict):
+        # first make a copy so we don't modify in place
```
marcomangano, line 3451: please add docstrings in this and the following methods

marcomangano (commented):
Nice stuff, most of my comments are to get a better understanding about what is going on.
Two comments:
- Shall we rename `convertSensitivitytoDict` and the reverse one? it seems it's of more general use than for just sensitivities
- I assume this is done by the new map functions at the end of `DVGeo.py`, but how are the DVs mapped back to the user at the end?

pygeo/constraints/thicknessConstraint.py
marcomangano: Yeah I also thought about `self.dtype` as a reason to cast `self.D0`. Will check the fixes when pushed

pygeo/parameterization/DVGeo.py
```diff
@@ -1237,6 +1239,30 @@ class in geo_utils. Using pointSelect discards everything in volList.
...
+        if u is not None:
+            # we are after a square matrix
+            if u.shape != (NDV, NDV):
+                raise ValueError(f"The shapes don't match! Got shape = {u.shape} but NDV = {NDV}")
+            if scale is None:
+                raise ValueError("If u is provided, then scale must also be provided.")
+            s = None
+        else:
+            if ptSetName is None:
+                raise ValueError("If u and s need to be computed, you must specify the ptSetName")
```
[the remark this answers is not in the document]
nwu63: This is needed if the user wants DVGeo to compute the necessary matrices.

pygeo/parameterization/DVGeo.py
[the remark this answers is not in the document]
nwu63: Yeah I agree, I'm not sure what's a good name but can definitely change that.

compName here is a boolean, and it can be False when useComposite is True, since we need to compute sensitivities both ways.

pygeo/parameterization/DVGeo.py
nwu63: The function `totalSensitivity` computes `dIdx`, which is the array of sensitivities wrt either the original DVs or the compDVs depending on useComposite. This function just converts that array to a dictionary populated by DV keys, and of course we just put one DVgroup in the dictionary when compDV is activated.

pygeo/parameterization/DVGeo.py
nwu63: Yeah, I can add a warning if ignoreVars is passed in.

pygeo/parameterization/DVGeo.py
anilyil: can you rename compName variable to something else? I suggest `useCompositeNames`. `comp` might be confused for "component", plus adding "use" makes it more clear that we expect a boolean instead of a list or a string

pygeo/parameterization/DVGeo.py
```diff
@@ -1237,6 +1239,30 @@ class in geo_utils. Using pointSelect discards everything in volList.
 
         return self.DV_listSectionLocal[dvName].nVal
 
+    def addGeoDVComposite(self, dvName, ptSetName=None, u=None, scale=None):
```
anilyil, line 1242: Please add some comments or dosctrings briefly explaining that this is more of a preprocessing routine thats run on a single proc.

anilyil (commented):
Good stuff. I have few minor comments, but good to go otherwise

marcomangano (approved):

anilyil (approved):

PR #127, opened 2022-03-23
Precise thickness and volume constraints for non-trapezoidal wings

sseraj:
## Purpose

For non-trapezoidal wings, we use `leList` and `teList` with length greater than 2 to define the locations between the root and tip. However, the points currently used in the thickness and volume constraints do not exactly match these intermediate locations. This is documented behavior in DVCon.

As an alternative, I added the option to pass a list for `nSpan` instead of an integer. Doing so will interpolate within each segment defined by `leList` and `teList` and match the intermediate locations. The behavior is unchanged if an integer is passed. I included an example of a cranked arrow wing below.

Thickness constraints with `nSpan=10`:
thickness_before

Thickness constraints with `nSpan=[5, 5]`:
thickness_after

The constraints now match the wing break location and follow the chordwise direction.

## Expected time until merged

One week

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing

I added tests for thickness and volume constraints with a list for `nSpan`.

## Checklist

### DIFF
--- a/pygeo/constraints/DVCon.py
+++ b/pygeo/constraints/DVCon.py
@@ -460,13 +460,14 @@ class DVConstraints:
 
         * The leading and trailing edges are approximated using
           2-order splines (line segments) and nSpan points are
-          interpolated in a linear fashion. Note that the thickness
-          constraint may not correspond **EXACT** to intermediate
-          locations in leList and teList. For example, in the example
-          above, with leList=3 and nSpan=3, the three thickness
+          interpolated in a linear fashion. For integer nSpan, the thickness
+          constraint may not correspond **EXACTLY** to intermediate
+          locations in leList and teList. In the example above,
+          with len(leList)=3 and nSpan=3, the three thickness
           constraints on the leading edge of the 2D domain would be at
           the left and right boundaries, and at the point denoted by
-          'o' which is equidistance between the root and tip.
+          'o' which is equidistant between the root and tip.
+          To match intermediate locations exactly, pass a list for nSpan.
 
         * If a curved leading or trailing edge domain is desired,
           simply pass in lists for leList and teList with a sufficient
@@ -490,18 +491,21 @@ class DVConstraints:
         leList : list or array
             A list or array of points (size should be (Nx3) where N is
             at least 2) defining the 'leading edge' or the start of the
-            domain
+            domain.
 
         teList : list or array
            Same as leList but for the trailing edge.
 
-        nSpan : int
+        nSpan : int or list of int
             The number of thickness constraints to be (linear)
-            interpolated *along* the leading and trailing edges
+            interpolated *along* the leading and trailing edges.
+            A list of length N-1 can be used to specify the number
+            for each segment defined by leList and teList and
+            precisely match intermediate locations.
 
         nChord : int
             The number of thickness constraints to be (linearly)
-            interpolated between the leading and trailing edges
+            interpolated between the leading and trailing edges.
 
         lower : float or array of size (nSpan x nChord)
             The lower bound for the constraint. A single float will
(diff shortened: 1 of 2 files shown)


pygeo/constraints/DVCon.py
```diff
@@ -3204,12 +3221,56 @@ def _generateIntersections(self, leList, teList, nSpan, nChord, surfaceName):
...
+        # Generate spanwise parametric distances
+        if isinstance(nSpan, int):
+            # Use equal spacing along the curve
+            le_span_s = te_span_s = np.linspace(0.0, 1.0, nSpan)
+        elif isinstance(nSpan, list):
+            # Use equal spacing within each segment defined by leList and teList
+
+            # We use the same nSpan for the leading and trailing edges, so check that the lists are the same size
+            if len(leList) != len(teList):
+                raise ValueError("leList and teList must be the same length if nSpan is provided as a list.")
```
marcomangano, line 3233: I know it's mentioned in the docstring above, but what are the instances where `leList` and `teList` would be intentionally inconsistent when `nSpan` is float?

marcomangano (approved):
Looks good! I just have one comment on the inputs, but it's not related to the specific additions of this PR

pygeo/constraints/DVCon.py
sseraj: That's a good question. I can't see it happening too often, but there could be some uses. For example, a wing could have a leading edge break and a straight trailing edge.

marcomangano: Oh yeah, that makes sense.

pygeo/constraints/DVCon.py
```diff
@@ -3204,12 +3221,56 @@ def _generateIntersections(self, leList, teList, nSpan, nChord, surfaceName):
...
+                # Interpolate over this segment and append to the parametric distance array
+                le_span_s = np.append(
+                    le_span_s, np.linspace(le_breakPoints[i], le_breakPoints[i + 1], nSpan[i], endpoint=endpoint)
+                )
+                te_span_s = np.append(
+                    te_span_s, np.linspace(te_breakPoints[i], te_breakPoints[i + 1], nSpan[i], endpoint=endpoint)
+                )
+
+            # nSpan needs to be the total number of sections for the remainder of the function
+            nSpan = sum(nSpan)
```
anilyil, line 3265: I am not sure if I like this change to what `nSpan` is mid-way through the routine. Initially its the list of spanwise section counts, but then it is switched to the total count. I guess it would be a bit better if we used a different variable for the total (or the span list)? I am not sure how useful my comment is but thought this type of using the same variable name for two things might be confusing?

anilyil (commented):
The changes look good. I only have one minor comment on the coding style regarding the different uses of the `nSpan` variable. I left it to one of the routines you added but obviously applies to both. Let me know what you think and we can either slightly modify it or I can just approve it if you don't want to make that change.

pygeo/constraints/DVCon.py
sseraj: I agree that the current approach is not very clean. I can propose two alternatives that would keep `nSpan` as an integer:
1) Add a separate kwarg that is a list defining the fraction of `nSpan` that gets allocated to each segment, e.g. `[0.5, 0.5]`.
2) Don't add any additional arguments. For a given uniform distribution in parametric space, 'snap' the closest point to the intermediate location and resample the curves. This would change the default behavior.

sseraj: Turns out Anil was referring to the variable names in the function, not the API. I will update the variable names to be clearer.

anilyil (approved):

marcomangano (approved):

PR #129, opened 2022-03-23
Fixed bug for composite DV

nwu63:
## Purpose
Apparently, I had hard-coded the initial DVs to zero. I didn't care at the time because in all my tests, the initial DVs happen to be zero anyways. But now I fixed it so it computes the correct initial DVs.

Also, I changed `getValues()` to return real values instead of complex. It didn't make sense to me to return any complex DVs, and I'm pretty sure the ESP/VSP versions will always return real values. The tests should all pass, and I'm somewhat confident in this change because `getValues()` is not really used internally in DVGeo. But maintainers please double check that this works.

## Expected time until merged
ASAP

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
N/A

## Checklist

### DIFF
--- a/pygeo/__init__.py
+++ b/pygeo/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "1.10.0"
+__version__ = "1.10.1"
 
 from .pyNetwork import pyNetwork
 from .pyGeo import pyGeo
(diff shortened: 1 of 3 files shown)


anilyil (dismissed):
the bug fix looks good. I like the `getValues` change but I am not sure if it would break anybody's workflow? Maybe some people are getting the values

anilyil (approved):
PR's so good I had to approve it twice

marcomangano (approved):

Issue #125, opened 2022-03-15
Define base classes to unify the various DVGeo modules


nwu63:
## Description
Currently, DVGeoVSP and DVGeoESP do not share any code with DVGeo, but mimics its API. This results in potential inconsistencies such as #124.

The solution is probably to define the API via an abstract base class, which is then inherited by all the various modules.

hajdik:
Closed with #130

PR #130, opened 2022-03-25
Base class for DVGeos

hajdik:
## Purpose
This is pretty much ready for review now, the biggest lingering questions are what the new classes should be named and if there are more functions that can be brought up a level after some refactoring. 

Addresses #125. There is repeated code and similar structures between the different `DVGeo` classes, resulting in issues when there are changes in the API. There is now an abstract base class, `BaseDVGeo`, that all `DVGeo` classes will inherit from, as well as an intermediate class, `DVGeoSketch`, for `DVGeoESP` and `DVGeoVSP` to inherit from due to the large amount of shared code. 

There is also a base class for design variables now to unify the different implementations better. 

## Expected time until merged
Maybe a week now that it's closer to complete, there's no rush

## Type of change
- [X] Refactoring (no functional changes, no API changes)
- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
I do not believe any additional tests will be needed, the functionality should be the same so if the original tests pass this should be good. 

## Checklist


sseraj (commented):
I think this is a good start. Some comments below

pygeo/parameterization/BaseDVGeo.py
```diff
@@ -0,0 +1,181 @@
...
+    @abstractmethod
+    def printDesignVariables(self, directory):
+        pass
+
+    @abstractmethod
+    def writePointSet(self, name, fileName):
+        pass
+
+    @abstractmethod
+    def demoDesignVars(self, directory):
```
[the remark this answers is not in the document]
hajdik: I have one for DVGeoESP on another branch, not sure how hard it would be to make a DVGeoVSP one too.

pygeo/parameterization/DVGeo.py
```diff
@@ -1830,6 +1831,7 @@ def applyToChild(self, iChild):
         child.refAxis.coef = child.coef.copy()
         child.refAxis._updateCurveCoef()
 
+    # TODO axi doesn't have this?
```
marcomangano, line 1834: Leftover comment

pygeo/parameterization/DVGeoSketch.py
```diff
@@ -0,0 +1,296 @@
...
+        for dvName in self.DVs:
+            dvDict[dvName] = self.DVs[dvName].value
+
+        return dvDict
+
+    @abstractmethod
+    def update(self, ptSetName, config=None):
+        pass
+
+    def pointSetUpToDate(self, ptSetName):
```
marcomangano, line 122: this function already exists in the parent class, why are you re-implementing it here?

pygeo/parameterization/BaseDVGeo.py
```diff
@@ -0,0 +1,226 @@
...
+        """
+        Return a list of the design variable names. This is typically used when specifying a wrt= argument for pyOptSparse.
+
+        Examples
+        --------
+        optProb.addCon(.....wrt=DVGeo.getVarNames())
+        """
+        pass
+
+    @abstractmethod
```
marcomangano, line 102: Do you still need to use \@abstractmethod here?

pygeo/parameterization/DVGeoSketch.py
```diff
@@ -0,0 +1,296 @@
...
+        ptSetName : str
+            The name of the pointset to check.
+        """
+        if ptSetName in self.updated:
+            return self.updated[ptSetName]
+        else:
+            return True
+
+    @abstractmethod
+    def getNDV(self):
```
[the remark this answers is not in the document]
hajdik: I tried that but the method in VSP returns the length of a different list than ESP does so using the same for both broke ESP tests so I just kept them separate instead of figuring out why they're different.

pygeo/parameterization/DVGeo.py
```diff
@@ -14,9 +14,10 @@
 import warnings
 from baseclasses.utils import Error
 from .designVars import geoDVGlobal, geoDVLocal, geoDVSpanwiseLocal, geoDVSectionLocal, geoDVComposite
+from .BaseDVGeo import BaseDVGeo
```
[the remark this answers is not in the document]
marcomangano: What about `BaseDVGeometry` to be consistent with `DVGeo.py`?

nwu63 (commented):
Overall I think there should be more things that we can pull out, but maybe that needs a bigger refactoring. See specific comments below

pygeo/parameterization/BaseDVGeo.py
```diff
@@ -0,0 +1,223 @@
...
+
+        Parameters
+        ----------
+        ptSetName : str
+            Name of point-set to return. This must match ones of the given in an :func:`addPointSet()` call.
+        """
+        pass
+
+    @abstractmethod
+    def writePointSet(self, name, fileName):
```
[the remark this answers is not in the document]
hajdik: It's called from demoDesignVars and also called on its own by the tutorial

pygeo/parameterization/BaseDVGeo.py
```diff
@@ -0,0 +1,223 @@
...
+
+        Parameters
+        ----------
+        name : str
+             The name of the point set to write to a file
+
+        fileName : str
+           Filename for tecplot file. Should have no extension, an extension will be added
+        """
+        coords = self.update(name, childDelta=True)
```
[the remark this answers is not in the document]
hajdik: Since the regular DVGeo has to have the child parameters I'll keep its writePointSet in its original state. I'll move this version (minus the childDelta) to the middle class for ESP and VSP to use and keep the top level class one as an abstract method.

sseraj:
The tests are hanging. I tried cancelling the build, but Azure is slow to respond. Testing locally first would be ideal

hajdik:
> 

They were passing locally, I'll look into it later today.

pygeo/parameterization/BaseDVGeo.py
```diff
@@ -0,0 +1,218 @@
...
+
+Holds a basic version of a DVGeo geometry object
+Enables the use of different geometry parameterizations (FFD, OpenVSP, ESP, etc) with the MACH-Aero framework
+"""
+
+from abc import abstractmethod
+from typing import OrderedDict
+
+
+class BaseDVGeometry:
```
joanibal, line 12: Does this need to be a ABC to work?

pygeo/parameterization/DVGeoSketch.py
```diff
@@ -0,0 +1,281 @@
...
+        self.projTol = projTol * self.meshScale  # default input is in meters.
+
+        self.updatedJac = {}
+        self.comm = comm
+
+        # Initial list of DVs
+        self.DVs = OrderedDict()
+
+    @abstractmethod
+    def addPointSet(self, points, ptName, **kwargs):
```
joanibal, line 79: Hmm, is there a way to inherent from the baseclass without redefining them here. 
I thought abstract methods only cause errors if you try to instantiate an object, so we should be fine to remove these abstract methods from this base class since it should never be used directly right?

joanibal:
```python
from abc import abstractmethod
from abc import ABC
 
class A(ABC):
    
    @abstractmethod
    def test(self):
        print('this shouldnt run...')
        pass 
    
class B(A):
    pass

class C(B):
    
    def test(self):
        print('hi')
        
        
myC = C()
myC.test()

myB = B()
myA = A()
```
I made a little test script to show you what I mean. 
It should fail at B() even though B has no virtual methods itself.

joanibal (commented):
I'm confused by some of the abstract class and methods stuff.

pygeo/parameterization/BaseDVGeo.py
nwu63: It should be `ABC`, but it doesn't have to be for it to work, that just prevents this class from being instantiated directly. The overloaded methods will still work without `ABC`.

pygeo/parameterization/DVGeoSketch.py
nwu63: In this case we can remove this I think, because
1. B is also an ABC so it will not be instantiated
2. A already defines `addPointSet` and this implementation does not add anything

pygeo/parameterization/DVGeoSketch.py
```diff
@@ -0,0 +1,281 @@
...
+        pass
+
+    @abstractmethod
+    def addVariable(self):
+        """
+        Add a design variable definition.
+        """
+        pass
+
+    def addVariablesPyOpt(self, optProb):
```
nwu63, line 216: As mentioned above, any empty function definition here which are already in `BaseDVGeo` can be removed. This is simply duplicated from that and doesn't serve any purpose.

pygeo/parameterization/BaseDVGeo.py
joanibal: But you make A an `ABC` then you get this error
```
Traceback (most recent call last):
  File "test.py", line 23, in <module>
    myA = A()
TypeError: Can't instantiate abstract class A with abstract methods test
```
Which seems more useful to me

nwu63: I see what you mean now, yes you are right. We should be using `ABC` here.

nwu63: Oh hold on I didn't catch this, I don't think `DVGeoMulti` should inherit anything since it encapsulates DVGeo objects.

pygeo/parameterization/DVGeo.py
```diff
@@ -169,7 +171,6 @@ def __init__(self, fileName, *args, isComplex=False, child=False, faceFreeze=Non
         self.links_n = None
 
         # Jacobians:
-        self.ptSetNames = []
```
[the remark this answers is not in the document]
hajdik: I believe all the shared attributes have been pulled out to the base class, the rest of the ones in here are for child DVGeos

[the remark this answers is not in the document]
hajdik: I believe all the shared attributes have been pulled out to the base class, the rest of the ones in here are for child DVGeos

pygeo/parameterization/DVGeoSketch.py
hajdik: Duplicated abstract methods have been removed from the intermediate class

pygeo/parameterization/BaseDVGeo.py
hajdik: DVGeoMulti has been set back to its original state before this PR.

nwu63 (approved):
This looks good to me, but I'd appreciate other diligent reviews

joanibal (approved):

PR #138, opened 2022-05-03
Fixing bug in local DVs

hajdik:
## Purpose
The changes in #130 to the DV classes introduced an issue that caused my cases and the tutorial wing and airfoil to not run, they exited immediately with SNOPT saying the problem was infeasible. I believe the issue is typos in `geoDVLocal` and `geoDVSpanwiseLocal`, which this should fix. 

The bigger question is why the MACH and pyGeo tests still passed with this issue and how tests could catch it.

## Expected time until merged
A day or two, there are only a few changed lines but someone who has cases that use different types of DVs might want to test with this to make sure everything is working.

## Type of change
- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
Without this change, running the tutorial scripts in Docker (I tried `u20-gcc-ompi-latest`) will result in an immediate exit with no apparent errors and a `10 14` code from SNOPT. 

## Checklist

### DIFF
--- a/pygeo/parameterization/designVars.py
+++ b/pygeo/parameterization/designVars.py
@@ -70,7 +70,7 @@ class geoDVLocal(geoDV):
 
         N = len(axis)
         nVal = len(coefList) * N
-        super().__init__(name=name, value=np.zeros(nVal, "D"), nVal=nVal * N, lower=lower, upper=lower, scale=scale)
+        super().__init__(name=name, value=np.zeros(nVal, "D"), nVal=nVal, lower=lower, upper=upper, scale=scale)
 
         self.config = config
         self.coefList = np.zeros((self.nVal, 2), "intc")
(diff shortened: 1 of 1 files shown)


pygeo/parameterization/designVars.py
```diff
@@ -229,12 +229,6 @@ def __init__(self, name, lower, upper, scale, axis, coefListIn, mask, config, se
         super().__init__(name=name, value=np.zeros(nVal, "D"), nVal=nVal, lower=None, upper=None, scale=scale)
 
         self.config = config
-        if lower is not None:
```
marcomangano, line 232: why are you removing these lines?

hajdik: That portion is duplicated in the class it inherits from, `geoDV`, so it shouldn't be necessary here.

joanibal (approved):

marcomangano (approved):

PR #140, opened 2022-05-06
Fixed initialization in base DV class

hajdik:
## Purpose
I found another issue I introduced when merging DVs into one base class in #130. Some parameters (`upper`, `lower`, `scale`) were not being set when a DV is created with their values as `None`, so there was an error when `addVarGroup` was called.  

## Expected time until merged
Should be quick, just one line and it fixes a bug

## Type of change
- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
Without this, some DVs will not run without bounds. 

## Checklist


sseraj:
We should consider testing the optimization related parts of the code.

nwu63:
Why didn't our MACH-Aero tests catch any of these?

marcomangano (approved):

joanibal (approved):

PR #141, opened 2022-06-01
CST airfoil parameterization

eytanadler:
## Purpose

This PR adds a CST parameterization for airfoil optimization. It allows for design variables of the CST parameters, class shape, and chord length. All derivatives analytically defined.

## Expected time until merged

2 furlongs

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing

Tests for the new `DVGeometryCST` can be found in `tests/reg_tests/test_DVGeometryCST.py`.

## Checklist


hajdik (commented):
Going to look at this more later, just pointing out a few places that can use the DVGeo and DV inheritance

pygeo/parameterization/DVGeoCST.py
```diff
@@ -0,0 +1,1013 @@
...
+            Index of the column in the point set to use as the vertical (y in CST) airfoil coordinates, by default 1
+        comm : MPI communicator, optional
+            Communicator for DVGeometryCST instance, by default MPI.COMM_WORLD
+        isComplex : bool, optional
+            Initialize variables to complex types where necessary, by default False
+        debug : bool, optional
+            Show plots when addPointSet is called to visually verify that it is correctly splitting
+            the upper and lower surfaces of the airfoil points, by default False
+        """
+        self.points = OrderedDict()  # For each point set, it contains a dictionary with the coordinates,
```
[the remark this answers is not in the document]
eytanadler: Also got rid of `self.updated`

hajdik (approved):
LGTM

nwu63:
Do we want to find another reviewer to expedite the process? Maybe @joanibal?

pygeo/parameterization/DVGeoCST.py
```diff
@@ -0,0 +1,1093 @@
...
+            This name will need to be provided when updating the coordinates or when
+            getting the derivatives of the coordinates.
+        kwargs
+            Any other parameters are ignored.
+        """
+        # Convert points to the type specified at initialization (with isComplex) and store the points
+        points = points.astype(self.dtype)
+
+        # Check that all points are within the airfoil x bounds
+        if np.any(points[:, self.xIdx] < self.xMin) or np.any(points[:, self.xIdx] > self.xMax):
```
anilyil, line 184: Shouldn't we allow small deviations? This can be a kwarg, and has to do with the accuracy of your mesh and how well it fits the airfoil coordinates.

pygeo/parameterization/DVGeoCST.py
```diff
@@ -0,0 +1,1093 @@
...
+            "n1_upper": False,
+            "n2_upper": False,
+            "n1_lower": False,
+            "n2_lower": False,
+            "chord": False,
+        }
+
+        # Default DVs to be copied for each point set
+        self.defaultDV = {
+            "upper": 0.1 * np.ones(8, dtype=self.dtype),
```
anilyil, line 106: Is this assuming we will always have 8 coefficients to control? If so, I think this should be an input argument at least. I am also confused on how this interacts with the DV addition. Is the user expected to stay below 8 per half airfoil?

pygeo/parameterization/DVGeoCST.py
```diff
@@ -0,0 +1,1093 @@
...
+            CST coefficients fit to the airfoil surface.
+        """
+        # Normalize x and y
+        chord = np.max(xCoord) - np.min(xCoord)
+        xCoord = (xCoord - np.min(xCoord)) / chord
+        yCoord /= chord
+
+        # Compute the coefficients via linear least squares
+        dydw = DVGeometryCST.computeCSTdydw(xCoord, N1, N2, np.ones(nCST), dtype=dtype)
+        w = np.linalg.lstsq(dydw.T, yCoord, rcond=None)[0]
```
anilyil, line 1049: any safety checks here or any metric we can report for the quality of the fit? Maybe its not even close due to some discrepancy with the input file.

anilyil (commented):
Sorry for the late review. Here are my comments after one pass. We can have a call in the next few days to discuss if needed.

pygeo/parameterization/DVGeoCST.py
```diff
@@ -0,0 +1,1093 @@
...
+        self.xMin = np.min(self.foilCoords[:, self.xIdx])
+        self.xMax = np.max(self.foilCoords[:, self.xIdx])
+
+        # Check that the leading edge is at y = 0
+        idxLE = np.argmin(self.foilCoords[:, self.xIdx])
+        yLE = self.foilCoords[idxLE, self.yIdx]
+        if abs(yLE) > 1e-2:
+            raise ValueError(f"Leading edge y (or idxVertical) value must equal zero, not {yLE}")
+
+        # Determine the trailing edge thickness
```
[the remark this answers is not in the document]
marcomangano: Haven't looked into the code here, but it sounds [prefoil](https://github.com/mdolab/prefoil) might have some of the machinery/logic that would help here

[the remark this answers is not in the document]
eytanadler: I think it would be nice to be able to support closed trailing edges (makes ADflow meshes easier), but I understand the problem with this approach. I'm open to implementing the angle detection approach. I don't see any tools to do this in prefoil but maybe I'm not looking hard enough; @marcomangano do you know where it'd be?

pygeo/parameterization/DVGeoCST.py
eytanadler: Good question. We do have a `debug` flag in the `__init__` and I could set it to plot the airfoil and fitted airfoil if that flag is set to True so we could at least visually compare. Thoughts?

pygeo/parameterization/DVGeoCST.py
```diff
@@ -0,0 +1,1093 @@
...
+        idxTE = np.where(self.foilCoords[:, self.xIdx] == self.xMax)[0]
+        self.yUpperTE = np.max(self.foilCoords[idxTE, self.yIdx])
+        self.yLowerTE = np.min(self.foilCoords[idxTE, self.yIdx])
+        self.thicknessTE = self.yUpperTE - self.yLowerTE
+
+        # Compute splines for the upper and lower surfaces (used to split the foil in addPointSet)
```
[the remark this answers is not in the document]
eytanadler: Good point, a subtle difference you found here that I didn't pick up on: prefoil splits with the LE as the point furthest from the TE, but  we also use the minimum x point for other operations (such as determining the chord). I don't think this is a problem though because there isn't overlap between where min x is used and where the prefoil split happens.

eytanadler:
Not sure what's going on with flake8, but it must be due to a new version since it's from code I didn't touch. What's the best approach to handle that?

nwu63:
> Not sure what's going on with flake8, but it must be due to a new version since it's from code I didn't touch. What's the best approach to handle that?

Yeah this is from `flake8-bugbear` which just had a new version released. It gets a bit tricky pinning versions of all packages but I guess we could do it. Though for this kind of thing new versions bring better bug detection.

pygeo/parameterization/DVGeoCST.py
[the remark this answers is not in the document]
marcomangano: [This method](https://github.com/mdolab/prefoil/blob/c59f087134517f2d4a9f1a621284da1f3abf7a67/prefoil/airfoil.py#L898) does the check that Anil is recommending AND opens up the airfoil. Not sure if it makes sense to have prefoil as a dependency for the cst module.

eytanadler:
> > Not sure what's going on with flake8, but it must be due to a new version since it's from code I didn't touch. What's the best approach to handle that?
> 
> Yeah this is from `flake8-bugbear` which just had a new version released. It gets a bit tricky pinning versions of all packages but I guess we could do it. Though for this kind of thing new versions bring better bug detection.

Got it. I'm happy to fix it in this PR, I'm just not exactly sure what it needs. Running flake8 locally with the MDO Lab config shows no errors. I tried adding a `nonlocal`, but that didn't do it. Maybe it needs the `nonlocal` and a new variable defined in the nested function from `nRefAxPts` to use a locally-defined variable in the for loop spec?

pygeo/parameterization/DVGeoCST.py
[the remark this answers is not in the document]
marcomangano: I think @akleb worked on that most recently, if I recall correctly the check works on `coords` which is the point cloud and not the spline (`self.spline.X` are the actual points, I believe) - which is just re-generated at the end . Anyway, the idea should hold and should be quick to re-implement. Good luck!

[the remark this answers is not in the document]
akleb: _\*gasp\*_ Did we just agree to code duplication in pygeo? Why not just pop the function that detects the trailing edge points out of the `removeTE` function into a new function and then have the `removeTE` function call it? Then you could just use the already written code. I don't really care either way, just seems like bad practice to rewrite code in different modules unless pygeo doesn't have a prefoil/pyspline dependency (the real argument for me would be if it didn't have a pyspline dependency (I know nothing about pygeo so idk what the case is)).

Also I am not really sure what is going on in this code, but perhaps there is a way to use more of prefoil if you're not already. I can talk with you about what it does, if you would like (I have no experience with DV objects).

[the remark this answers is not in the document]
eytanadler: Anil and I are meeting later today to figure out the best approach since it is complicated by the fact that the trailing edge needs to be separated out for each partitioned group of points in the point set individually, rather than all airfoil coordinates at once. We'll keep prefoil in mind, thanks for the suggestions.

nwu63:
Yeah at first glance I'm not really sure either. If we can't figure it out we can always tell flake8 to skip the warning.

pygeo/parameterization/DVGeoCST.py
```diff
@@ -0,0 +1,1175 @@
...
+        # preFoil defines the leading edge as the point furthest from the trailing edge
+        self.foil = Airfoil(coords)
+        self.upperSpline, self.lowerSpline = self.foil.splitAirfoil()
+
+        # Fit CST parameters to the airfoil's upper and lower surface
+        self.idxFoil = {}
+        self.idxFoil["upper"], self.idxFoil["lower"] = self._splitUpperLower(self.foilCoords)
+        chord = self.xMax - self.xMin
+        self.defaultDV["chord"][0] = chord
+        print(f"######## Fitting CST coefficients to coordinates in {datFile} ########")
```
anilyil, line 223: is this printed on all procs? Also related; is the fit done in parallel on all procs or do you do it on once and communicate the result? if you are doing it in parallel, there's a small chance the procs will get different values, especially when the fit is ill conditioned

pygeo/parameterization/DVGeoCST.py
```diff
@@ -0,0 +1,1175 @@
...
+                xPts / chord,
+                self.defaultDV["n1_lower"],
+                self.defaultDV["n2_lower"],
+                self.defaultDV[dvType],
+                yTE,
+                dtype=self.dtype,
+            )
+            L2norm = np.sqrt(1 / ptsFit.size * np.sum((self.foilCoords[self.idxFoil[dvType], self.yIdx] - ptsFit) ** 2))
+
+            print(f"{dvType.capitalize()} surface")
```
anilyil, line 247: Again, not all procs should print

pygeo/parameterization/DVGeoCST.py
```diff
@@ -0,0 +1,1175 @@
...
+            "xMax": self.xMax.copy(),
+            "xMin": self.xMin.copy(),
+            "thicknessTE": self.thicknessTE.copy(),
+        }
+
+        # Determine which points are on the upper and lower surfaces
+        self.points[ptName]["upper"], self.points[ptName]["lower"] = self._splitUpperLower(points)
+
+        # If debug mode is on, plot the upper and lower surface points
+        if self.debug:
```
anilyil, line 292: similar to the print, it might be better to do this while keeping in mind that the run can be parallel. maybe we can communicate everything back to root and do the plot once. it gets tricky because dvcon adds the same points on all procs whereas adflow will add a distributed pointset

pygeo/parameterization/DVGeoCST.py
```diff
@@ -0,0 +1,1175 @@
...
+                self.points[ptName]["points"][:, self.yIdx],
+                s=3,
+                c="k",
+                zorder=3,
+            )
+            plt.legend(["Upper", "Lower"])
+            plt.show()
+            plt.close(fig)
+
+    def addDV(self, dvName, dvType, lower=None, upper=None, scale=1.0, default=None):
```
anilyil, line 315: I just realized this; isnt it confusing that the DV names can get upper and lower and there are also upper and lower bounds on all dvs? maybe we change the upper and lower dvs to upper_coef and lower_coef? any name other than the upper/lower naming convention is probably fine

pygeo/parameterization/DVGeoCST.py
```diff
@@ -0,0 +1,1175 @@
...
+        -------
+        nDV : int
+            Total number of design variables
+        """
+        nDV = 0
+        for DV in self.DVs.values():
+            nDV += DV.nVal
+        return nDV
+
+    def printDesignVariables(self):
```
anilyil, line 808: do other dvgeos have this? its a good idea, but maybe just print from root?

pygeo/parameterization/DVGeoCST.py
anilyil: you print the L2 norm above so I am marking this as resolved.

pygeo/parameterization/DVGeoCST.py
```diff
@@ -0,0 +1,1175 @@
...
+            Second class shape parameter to assume in fitting, by default 1.0
+        dtype : type, optional
+            Type for instantiated arrays, by default float
+
+        Returns
+        -------
+        np.ndarray (nCST,)
+            CST coefficients fit to the airfoil surface.
+        """
+        # Normalize x and y
```
anilyil, line 1124: currently this fit is done in parallel on all procs. this is probably fine, but then we do have cases where the fit is ill conditioned, and it can likely lead to different fits on different procs. maybe its better to run this fit only on root and bcast w in the comm to other procs

anilyil (commented):
Looking a lot better. few issues left

anilyil (approved):
Looks good. Thanks for working on the changes

eytanadler:
> Looks good. Thanks for working on the changes

Thanks! Should I bump the version too? Minor?

hajdik (approved):

PR #144, opened 2022-07-11
Adding in child FFD functions

bernardopacini:
## Purpose
Adding a method for using parent + child FFDs in MPHYS. This is one possible implementation that I have been using with MPHYS + DAFoam and works as expected, but I think we should discuss if this is the _best_ approach for how to do this. The tricky part is in the order of operations: child FFDs cannot be added to a parent without a reference axis, so they cannot be immediately set to the MPHYS geometry object. The configure method then goes in the wrong order, so we need an explicit call to `nom_add_children()` (added in this PR) to actually add the child FFDs after the reference axes are declared but before any dependent constraints are defined. The better solution here might require more careful modifications to pyGeo, but I am not sure what other issues we might run into if we were to do that.

EDIT: PR #146 broke the MPHYS wrapper that was using deprecated calls to add local and global design variables. I fixed that here, though that will require updating the MPHYS examples and tests on the MPHYS repo too.

## Expected time until merged
~This PR is open to discussion -- no specific timeframe set.~ 2 weeks.

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
N / A

## Checklist


bernardopacini:
Part of the formatting failure is with one of the regression tests (unrelated to this PR). I do not think there is an actual issue, so I am just going to ignore it with a `#noqa:` statement.

```
./tests/reg_tests/test_DVGeometryMulti.py:103:35: B023 Function definition does not bind loop variable 'nRefAxPts'.
                for i in range(1, nRefAxPts)
```

joanibal (approved):
Looks good to me. I think we should look at pyGeo as discussed to see why the children FFD need reference axis before they are added.

anilyil:
The code looks good. my only comment is that wouldn't it make more sense to use a dictionary approach for the child FFDs? That way we can index them by name instead of order of addition. We can use an ordered dictionary under the hood, or sort the names alphabetically and keep track of them that way. the users can pass dictionaries with child ffd name as the key and file as the value. Let me know if you think this is useful. I am also fine with merging this PR as is and doing that change separately

bernardopacini:
> The code looks good. my only comment is that wouldn't it make more sense to use a dictionary approach for the child FFDs? That way we can index them by name instead of order of addition. We can use an ordered dictionary under the hood, or sort the names alphabetically and keep track of them that way. the users can pass dictionaries with child ffd name as the key and file as the value. Let me know if you think this is useful. I am also fine with merging this PR as is and doing that change separately

That approach makes sense to me. I went with this approach because it matches how pyGeo handles children, but if the indexing is confusing we could add a dictionary. If @joanibal is on board too I'll convert it to using dictionaries instead.

anilyil (dismissed):

bernardopacini:
@joanibal and I found that with a small modification to pyGeo we can actually simplify adding child FFDs in MPHYS. Hold off on merging this PR until we finalize that.

bernardopacini:
I fixed the package import in `BaseDVGeo.py` from:

```
from typing import OrderedDict
```

to 

```
from collections import OrderedDict
```
as we discussed offline.

This is now good to go!

joanibal (approved):

ArshSaja (approved):

PR #145, opened 2022-07-19
Extending DVGeoVSP to work with internal volume points

timryanb:
## Purpose
Slight modification to DVGeoVSP to allow for projection of points within the volume of a VSP body.
This should be more general than the original surface projection method and will allow for geometries with internal members (like structural models) to be parametrized using DVGeoVSP. With this PR, the more general volume projection procedure will become the new default procedure for DVGeoVSP.

## Expected time until merged
a week

## Type of change

## Testing
Any previously working DVGeoVSP model should be sufficient to verify that things are working.
Updated the previous DVGeoVSP regression test to work with this new formulation.

## Checklist


anilyil:
We need to update the docker images for CI to work because the volume projection requires the latest VSP version. I will create a PR for this.

pygeo/mphys/mphys_dvgeo.py
```diff
@@ -15,7 +15,7 @@ def initialize(self):
 
         self.options.declare("ffd_file", default=None)
         self.options.declare("vsp_file", default=None)
-        self.options.declare("vsp_options", default=None)
+        self.options.declare("vsp_options", default={})
```
joanibal, line 18: I think we shouldn't make that default argument `{}`.  Since it is a mutable type, if we have more than one DVGeoVSP objects by default the `vsp_options` objects will be linked. 
Perhaps we could make the option required and add in the description that it is dictionary of options and remove the default altogether.

joanibal (changes requested):
Looks good. 
Just one small request.

pygeo/mphys/mphys_dvgeo.py
timryanb: Agreed. I changed the default to a None. Later in the setup procedure (https://github.com/timryanb/pygeo/blob/vsp-volume-proj/pygeo/mphys/mphys_dvgeo.py#L27-L34), I check if the value is None and if so pass a unique empty dict for each DVGeo instance

anilyil:
I am working on updating the docker images to test this PR now. will keep you posted

joanibal (approved):

anilyil (commented):
The changes look good. I only have one comment: The changes between lines 250 and 259 are a bit confusing to me. I did not really understand the logic behind how sg is set. Also, the tg is set to 0.5. Does this cause any performance loss with purely surface based geometries? 

In general, my concerns raise from not understanding how the volume parameterization is done. It looks like the code does work. If that projection initial guess does not affect performance too much, I am fine with merging it. Even if it does affect performance, there's a failsafe projection w/o the initial guess so that is fine. 

@timryanb, it would help a bit  if you can include some comments on how the volume parameterization is done for closed volumes and for purely surface based geometries, and how these relate to the projection and the provided initial guesses. Like I said, the code changes are probably fine but a bit more explanation would be nice.

Regarding testing this on the docker images, we have the PR up, but we ended up needing additional changes to the u18 images. @nwu63 was planning on moving the u18 images to u22, so we did not look into updating the cmake version on the u18 images. If it looks like the u22 migration will take longer, we will try to update the cmake version. The latest VSP installed fine on the u20 images. 

Once the tests pass with the new image, and we have a bit of explanation of the volume projection, I am fine with merging the PR. Thanks for the work @timryanb!

timryanb:
> The changes look good. I only have one comment: The changes between lines 250 and 259 are a bit confusing to me. I did not really understand the logic behind how sg is set. Also, the tg is set to 0.5. Does this cause any performance loss with purely surface based geometries?
> 
> In general, my concerns raise from not understanding how the volume parameterization is done. It looks like the code does work. If that projection initial guess does not affect performance too much, I am fine with merging it. Even if it does affect performance, there's a failsafe projection w/o the initial guess so that is fine.
> 
> @timryanb, it would help a bit if you can include some comments on how the volume parameterization is done for closed volumes and for purely surface based geometries, and how these relate to the projection and the provided initial guesses. Like I said, the code changes are probably fine but a bit more explanation would be nice.
> 
> Regarding testing this on the docker images, we have the PR up, but we ended up needing additional changes to the u18 images. @nwu63 was planning on moving the u18 images to u22, so we did not look into updating the cmake version on the u18 images. If it looks like the u22 migration will take longer, we will try to update the cmake version. The latest VSP installed fine on the u20 images.
> 
> Once the tests pass with the new image, and we have a bit of explanation of the volume projection, I am fine with merging the PR. Thanks for the work @timryanb!

@anilyil, updated the code with more comments explaining volume projection procedure.

From my experience the performance from going from the surface to volume projection procedure has been negligible in compute time and equivalent in robustness. @lamkina, has also confirmed this on his engine cases.

anilyil (approved):
Thanks for the comments! The VSP update is merged to the docker repo and the new images will be available tomorrow. Once they are available, we can run the tests for this PR and merge it.

hajdik:
I think the new tests haven't ran yet because of a merge conflict in `DVGeoVSP.py`, once that is resolved the tests on the new docker image should run.

joanibal (approved):

anilyil (approved):

PR #150, opened 2022-08-02
Make ref axis optional when adding a child FFD

joanibal:
## Purpose
This PR addresses some challenges @bernardopacini was having when using child FFDs with MPhys. 

## Expected time until merged
3 days (@bernardopacini is waiting on this)

## Type of change
- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
I ran the tandem wing optimization tutorial problem, which has 2 child FFDs. I added a ref axis before adding one the parent and the other after. The results of the optimization are exactly the same.
## Checklist


pygeo/parameterization/DVGeo.py
```diff
@@ -1672,6 +1673,15 @@ def update(self, ptSetName, childDelta=True, config=None):
         # We've postponed things as long as we can...do the finalization.
         self._finalize()
 
+        for iChild in range(len(self.children)):
```
bernardopacini, line 1676: Should we be adding this before we finalize the parent to ensure it has all of the required point-sets embedded?

joanibal: I don't think it matters since `_finalize` only computes the links for the FFD control points not the embed points.

bernardopacini: Got it. That makes sense! I misinterpreted what was happening there.

pygeo/parameterization/DVGeo.py
```diff
@@ -1685,6 +1686,15 @@ def update(self, ptSetName, childDelta=True, config=None):
...
+            if len(self.children[iChild].axis) > 0:
+                self.children[iChild]._finalize()
```
marcomangano, line 1691: As you mentioned, the `_finalize` here is needed to get the `refAxis.coef` (coefficient of refAxis nodes) that are then embedded in the parent control volume below.
I am wondering then:
1. if it still makes sense to have another `child._finalize()` [here](https://github.com/mdolab/pygeo/blob/23150ad5f348f1574612e68ea43ff68277424173/pygeo/parameterization/DVGeo.py#L1780)
2. what caused the mismatch in the previous tests - was it due to the fact that the control points themselves were not embedded or something? I am honestly a bit confused by how pyGeo currently works, maybe there are some other intermediate steps we are missing if the axis is not added rightaway

marcomangano (commented):
This looks alright, the current scripts should not even be affected except for the "anticipated" `_finalize()` call I mention below.
I would still like to have the original developers feedback on this though

pygeo/parameterization/DVGeo.py
```diff
@@ -1685,6 +1686,15 @@ def update(self, ptSetName, childDelta=True, config=None):
...
+                refaxis_ptSetName = "child%d_axis" % (iChild)
+                if refaxis_ptSetName not in self.FFD.embeddedVolumes:
```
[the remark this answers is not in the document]
joanibal: exactly

pygeo/parameterization/DVGeo.py
joanibal: 1. The finalize method exists right away if it has already been called. This it is safe to repeatedly call finalize during each call of the update method.

joanibal: 2. I don't think I know which tests you are referring to.

marcomangano: I was referring to the test failures after the first commit, but the build log is lost now.
Agreed on the multiple finalizing calls.

pygeo/parameterization/DVGeo.py
```diff
@@ -1686,22 +1686,22 @@ def update(self, ptSetName, childDelta=True, config=None):
...
-
         # Make sure coefficients are complex
         self._complexifyCoef()
 
         # Set all coef Values back to initial values
         if not self.isChild:
             self.FFD.coef = self.origFFDCoef.copy()
             self._setInitialValues()
+            
+            for iChild in range(len(self.children)):
```
marcomangano, line 1697: Good catch here, did not realize that the children were trying to initialize themselves again. Out of curiosity, did you have issues because of this or you just re-checked the code?

marcomangano (commented):
Good fix, can merge after the code is updated with main and the style errors are fixed

bernardopacini:
> An optional thought; recently I realized that `def nom_addRefAxis(self,comp=None, **kwargs):` can be removed from `mphys_dvgeo.py` since it is easy to call the function from group level which is smth similar to this: `self.geo.DVGeo.addRefAxis(name="wing",curve=c1,volumes=[0] )`

I agree with this point, though I think we should leave it for a separate PR and discuss it in the MPHYS meeting. Does that seem reasonable? It ultimately still needs the child FFD to be added to the parent to be able to add the axis as you are suggesting, so it would rely on Josh's fix here.

bernardopacini (approved):
Once the formatting issues are fixed, looks good to me!

ArshSaja:
Agree! we can talk about it later.

ArshSaja (dismissed):

bernardopacini (approved):

marcomangano:
@joanibal: @bernardopacini and I just discussed about this PR and figured that this is probably not the best approach to address the MPhys issues. The fix in #144 already makes sense and covers MPhys problems. 
The fix in this PR is alright but 1) we are adding a limitation on the number of children levels you can add (not the best practice, but I think now we can have grandchildren, grand-grandchildren and so on) and 2) I feel that instead of adding and additional patch, we should sit down, understand the full workflow, and possibly refactor the FFD initialization and update sequence.

bernardopacini (approved):
With the added error message, this looks good to me.

ArshSaja (approved):

PR #155, opened 2022-09-06
Adding ESP into MPhys wrapper

hajdik:
## Purpose
@bernardopacini  and I added functions needed for `DVGeoESP` into the MPhys wrapper. Checks were also added to ensure the correct DVGeo type has been instantiated when using wrapper functions that are only for specific types, which fixes  the `DVGeoVSP` part of the wrapper because the previous addition of child FFD functions assumed all DVGeos can use child FFDs. 

## Expected time until merged
1 week unless we find something else this breaks

## Type of change
- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
This works with DAFoam and ADflow ESP-MPhys scripts. If I have time I'll add a test

## Checklist

### DIFF
--- a/pygeo/mphys/mphys_dvgeo.py
+++ b/pygeo/mphys/mphys_dvgeo.py
@@ -21,41 +21,31 @@ import os
 class OM_DVGEOCOMP(om.ExplicitComponent):
     def initialize(self):
 
-        self.options.declare("ffd_file", default=None)
-        self.options.declare("vsp_file", default=None)
-        self.options.declare("esp_file", default=None)
-        self.options.declare("vsp_options", default=None)
-        self.options.declare("esp_options", default=None)
+        self.options.declare("geo_file", default=None)
+        self.options.declare("geo_type", default=None)
+        self.options.declare("geo_options", default=None)
 
     def setup(self):
-        self.geo_type = None
-
         # create the DVGeo object that does the computations
-        if self.options["ffd_file"] is not None:
+        if self.options["geo_type"] == "ffd":
             # we are doing an FFD-based DVGeo
-            ffd_file = self.options["ffd_file"]
-            self.DVGeo = DVGeometry(ffd_file)
-            self.geo_type = "ffd"
+            self.DVGeo = DVGeometry(self.options["geo_file"])
 
-        if self.options["vsp_file"] is not None:
+        elif self.options["geo_type"] == "vsp":
             # we are doing a VSP-based DVGeo
-            vsp_file = self.options["vsp_file"]
-            if self.options["vsp_options"] is None:
+            if self.options["geo_options"] is None:
                 vsp_options = {}
             else:
-                vsp_options = self.options["vsp_options"]
-            self.DVGeo = DVGeometryVSP(vsp_file, comm=self.comm, **vsp_options)
-            self.geo_type = "vsp"
+                vsp_options = self.options["geo_options"]
+            self.DVGeo = DVGeometryVSP(self.options["geo_file"], comm=self.comm, **vsp_options)
 
-        if self.options["esp_file"] is not None:
+        elif self.options["geo_type"] == "esp":
             # we are doing an ESP-based DVGeo
-            esp_file = self.options["esp_file"]
             if self.options["esp_options"] is None:
                 esp_options = {}
             else:
-                esp_options = self.options["esp_options"]
-            self.DVGeo = DVGeometryESP(esp_file, comm=self.comm, **esp_options)
-            self.geo_type = "esp"
+                esp_options = self.options["geo_options"]
+            self.DVGeo = DVGeometryESP(self.options["geo_file"], comm=self.comm, **esp_options)
 
         self.DVCon = DVConstraints()
         self.DVCon.setDVGeo(self.DVGeo)
--- a/pygeo/mphys/mphys_dvgeo.py
+++ b/pygeo/mphys/mphys_dvgeo.py
@@ -20,35 +20,40 @@ import numpy as np
 class OM_DVGEOCOMP(om.ExplicitComponent):
     def initialize(self):
 
-        self.options.declare("geo_file", default=None)
-        self.options.declare("geo_type", default=None)
-        self.options.declare("geo_options", default=None)
+        self.options.declare("file", default=None)
+        self.options.declare("type", default=None)
+        self.options.declare("options", default=None)
 
     def setup(self):
-        self.geo_type = self.options["geo_type"]
+        self.type = self.options["type"]
 
         # create the DVGeo object that does the computations
-        if self.geo_type == "ffd":
+        if self.type == "ffd":
             # we are doing an FFD-based DVGeo
-            self.DVGeo = DVGeometry(self.options["geo_file"])
+            if self.options["options"] is None:
+                ffd_options = {}
+            else:
+                ffd_options = self.options["options"]
 
-        elif self.geo_type == "vsp":
+            self.DVGeo = DVGeometry(self.options["file"], comm=self.comm, **ffd_options)
+
+        elif self.type == "vsp":
             # we are doing a VSP-based DVGeo
-            if self.options["geo_options"] is None:
+            if self.options["options"] is None:
                 vsp_options = {}
             else:
-                vsp_options = self.options["geo_options"]
+                vsp_options = self.options["options"]
 
-            self.DVGeo = DVGeometryVSP(self.options["geo_file"], comm=self.comm, **vsp_options)
+            self.DVGeo = DVGeometryVSP(self.options["file"], comm=self.comm, **vsp_options)
 
-        elif self.geo_type == "esp":
+        elif self.type == "esp":
             # we are doing an ESP-based DVGeo
-            if self.options["geo_options"] is None:
+            if self.options["options"] is None:
                 esp_options = {}
             else:
-                esp_options = self.options["geo_options"]
+                esp_options = self.options["options"]
 
-            self.DVGeo = DVGeometryESP(self.options["geo_file"], comm=self.comm, **esp_options)
+            self.DVGeo = DVGeometryESP(self.options["file"], comm=self.comm, **esp_options)
 
         self.DVCon = DVConstraints()
         self.DVCon.setDVGeo(self.DVGeo)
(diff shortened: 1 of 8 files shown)


anilyil (commented):
The changes look good. One comment is on the initialization. Maybe we can modify the input arguments, have a single file input, and then determine the geo type based on the file extension? This would be a bit more foolproof. It's a bit tricky since I feel people use different extensions for the FFD files. VSP ones are all .vsp3 I think. I dont know the ESP extension.

hajdik:
> Maybe we can modify the input arguments, have a single file input, and then determine the geo type based on the file extension?  

That would work for ESP, they should all be `.csm`

> It's a bit tricky since I feel people use different extensions for the FFD files. 

Is there a set of allowable extensions for FFDs we could look for? Or if `.vsp3` or `.csm` aren't used, it could default to using FFD.

bernardopacini:
> > Maybe we can modify the input arguments, have a single file input, and then determine the geo type based on the file extension?  
> 
> 
> 
> That would work for ESP, they should all be `.csm`
> 
> 
> 
> > It's a bit tricky since I feel people use different extensions for the FFD files. 
> 
> 
> 
> Is there a set of allowable extensions for FFDs we could look for? Or if `.vsp3` or `.csm` aren't used, it could default to using FFD.

This makes me a bit nervous because even though the extension identifies the file, it isn't technically _necessary_ to read in the file (in any format, I believe). A user might write a file without an extension and then not understand why it doesn't load into pyGeo.

Would having an option for "geo_type" be a reasonable compromise?

anilyil:
Yeah, explicitly selecting geo_type also sounds good to me. Just a bit more explicit. Final decision is up to you, just raised that minor issue. I can approve the PR as is now as well.

bernardopacini:
> Yeah, explicitly selecting geo_type also sounds good to me. Just a bit more explicit. Final decision is up to you, just raised that minor issue. I can approve the PR as is now as well. 

Sounds good to me, if it's okay with @hajdik. I like the idea of having a single option for file path / name and then specifying the geo type. Thanks for bringing this up!

hajdik:
That sounds good to me, I can add that option in.

bernardopacini:
> That sounds good to me, I can add that option in.

On that note, could we also just make an “options” option instead of having the ESP, VSP, and FFD ones separate?

anilyil:
I would prefer just using individual openmdao options for each option instead of one option taking in a dictionary. Is that what you meant?

bernardopacini:
> I would prefer just using individual openmdao options for each option instead of one option taking in a dictionary. Is that what you meant?

The options are different for each parametrization type, so I think we'd end up with a lot of options (and many combinations that are incompatible). I do like the idea of trying to map them directly without having a big dictionary as we have now, but I worry that might end up being many, many options listed in the wrapper?

anilyil:
Now I see what you mean. so you are referring to the additional (optional) input arguments to initialize different classes? I am fine with adding that as a single option dictionary.

bernardopacini:
Yes, exactly. We should combine these:

```
       self.options.declare("vsp_options", default=None)
       self.options.declare("esp_options", default=None)
```

anilyil (dismissed):

pygeo/mphys/mphys_dvgeo.py
```diff
@@ -6,39 +6,52 @@
...
 
 # class that actually calls the dvgeometry methods
 class OM_DVGEOCOMP(om.ExplicitComponent):
     def initialize(self):
 
-        self.options.declare("ffd_file", default=None)
-        self.options.declare("vsp_file", default=None)
-        self.options.declare("vsp_options", default=None)
+        self.options.declare("geo_file", default=None)
+        self.options.declare("geo_type", default=None)
```
bernardopacini, line 24: I see later in the code there are still references to `self.geo_type`. Is this declared option here automatically converted to a member variable or do we need to do that ourselves and set `self.geo_type = self.options["geo_type"]`?

anilyil: I'd say saving it as self.geo_type is fine because otherwise you go through one attribute and one dictionary.

anilyil: the option is not converted to attribute automatically, we have to save it. sorry I didnt read the question fully. we can save it when we read the input parameters.

bernardopacini: Sounds good! I think the code as it is is broken since we don't save the option saved. Once that is saved it should be fixed.

anilyil (approved):
Thanks for the fixes.

pygeo/parameterization/DVGeoESP.py
```diff
@@ -245,6 +245,7 @@ def addPointSet(self, points, ptName, distributed=True, cache_projections=False,
         """
 
         # save this name so that we can zero out the jacobians properly
+        self.ptSetNames.append(ptName)
```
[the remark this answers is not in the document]
hajdik: That's in line with what the FFD `DVGeo` does so I think it's fine.

pygeo/mphys/mphys_dvgeo.py
```diff
@@ -6,39 +6,55 @@
...
+        elif self.options["geo_type"] == "esp":
+            # we are doing an ESP-based DVGeo
+            if self.options["geo_options"] is None:
+                esp_options = {}
+            else:
+                esp_options = self.options["geo_options"]
+
+            self.DVGeo = DVGeometryESP(self.options["geo_file"], comm=self.comm, **esp_options)
 
+        self.geo_type = self.options["geo_type"]
```
marcomangano, line 51: I might be picky, but maybe this could be done before the if clause?

pygeo/mphys/mphys_dvgeo.py
```diff
@@ -131,6 +162,9 @@ def nom_addLocalDV(self, dvName, axis="y", pointSelect=None, childIdx=None):
         return nVal
 
     def nom_addVSPVariable(self, component, group, parm, **kwargs):
+        # VSP DVs are only added to VSP-based DVGeo objects
+        if self.geo_type != "vsp":
+            raise RuntimeError(f"Only VSP-based DVGeo objects can use local DVs, not type:{self.geo_type}")
```
marcomangano, line 167: `local DVs` --> `VSPvariable` ?

pygeo/mphys/mphys_dvgeo.py
```diff
@@ -144,6 +178,20 @@ def nom_addVSPVariable(self, component, group, parm, **kwargs):
         # add the input with the correct value, VSP DVs always have a size of 1
         self.add_input(dvName, distributed=False, shape=1, val=val)
 
+    def nom_addESPVariable(self, desmptr_name, **kwargs):
+        # ESP DVs are only added to VSP-based DVGeo objects
+        if self.geo_type != "esp":
+            raise RuntimeError(f"Only ESP-based DVGeo objects can use local DVs, not type:{self.geo_type}")
```
marcomangano, line 184: `local DVs` --> `ESPvariable` ?

marcomangano (commented):
Just some minor comments on the code.
Are you sure you just want to pass in the options for the different kinds of parametrizations as a dict (?), without any check before initializing the class? I assume you can rely on error messages when `__init__` is run, but could be confusing in the long term imho

bernardopacini:
> Just some minor comments on the code.
> Are you sure you just want to pass in the options for the different kinds of parametrizations as a dict (?), without any check before initializing the class? I assume you can rely on error messages when `__init__` is run, but could be confusing in the long term imho

Regardless of if they are passed as dicts corresponding to the parametrization or as a general dict, there is no actual check for the options lining up in MPhys. So, the user could make the mistake regardless.

A proper fix for this would be to check in the `__init__` function of the respective parametrization if that option in the dictionary is part of the pyGeo object. One approach is by popping the kwarg key each time a value is used and checking if the dictionary is empty at the end of the function. @hajdik is this something you'd be willing to implement or would you like me to?

marcomangano:
> Edit: this opens a fairly large can of worms as we really should be doing this for _all_ our `__init__` functions. Maybe it would be worth adding this as a separate PR since the single dictionary proposed here is not any less safe than the previous MPhys implementation?

Agreed, and as I mentioned the error messages should be already clear enough for the user

marcomangano (dismissed):

hajdik:
> Maybe it would be worth adding this as a separate PR since the single dictionary proposed here is not any less safe than the previous MPhys implementation?

I think that should be a separate PR with wider discussion on the DVGeo modules themselves and this can stay as just touching the MPhys wrapper with the bugfixes for VSP + adding in ESP.

bernardopacini:
> > Edit: this opens a fairly large can of worms as we really should be doing this for _all_ our `__init__` functions. Maybe it would be worth adding this as a separate PR since the single dictionary proposed here is not any less safe than the previous MPhys implementation?
> 
> Agreed, and as I mentioned the error messages should be already clear enough for the user

I think there is a chance that a user specifies an option that is invalid and does not receive an error message. But, this could happen whenever instantiating pyGeo so this MPhys version is not any worse than before. I certainly agree with you that it is an issue we should address.

bernardopacini:
This is minor, but while we are making backwards incompatible changes: can we remove `geo_` from the options? Since it is the pyGeo object it should already be clear that this pertains to geometry. Then we would just have `file`, `type`, `options`. What do you all think?

pygeo/mphys/mphys_dvgeo.py
```diff
@@ -6,39 +6,56 @@
...
-        if self.options["ffd_file"] is not None:
+        if self.geo_type == "ffd":
             # we are doing an FFD-based DVGeo
-            ffd_file = self.options["ffd_file"]
-            self.DVGeo = DVGeometry(ffd_file)
-        if self.options["vsp_file"] is not None:
-            # we are doing a VSP based DVGeo
-            vsp_file = self.options["vsp_file"]
-            if self.options["vsp_options"] is None:
+            self.DVGeo = DVGeometry(self.options["geo_file"])
```
bernardopacini, line 33: Also, does the FFD version not have any options that would be passed through MPhys as VSP and ESP do?

bernardopacini (approved):

PR #168, opened 2022-11-30
Adding custom reference axis projections

anilyil:
## Purpose

The `axis` parameter when adding a reference axis determines how the links are created between each control point and the reference axis. The previous behavior only created rays from the control points and found the closest point between the ray and the reference axis to create the link from the control point to the reference axis.

This approach cannot guarantee that the links will lie on the same plane a section of the FFD cross section lies. To demonstrate this, consider a wing FFD that also has some dihedral built into it. The default axis direction is `x` so the code creates rays in the x direction from all control points. Then it finds the closest point from the ref axis to the ray. Due to the dihedral, the points on the upper surface will map slightly outboard, while the points on the lower surface will map inboard. See figure that demonstrates this default behavior:

demo_dihedral_bad

In this PR, I add the capability to create the links based on the intersection of a plane and the reference axis. The plane for each control point is defined by the control point itself and a user supplied normal. The normal is handled through the axis parameter, and the code determines behavior based on the type of the input. 

With the example above, if we set the axis parameter to `np.array([0.0, 0.0, 1.0)]`, the code creates a plane for each control point with the normal pointing in the z direction. Then, the reference axis links are created using the intersection of these planes with the reference axis. As a result, the control point links lie in the same plane at each section as shown here:

demo_dihedral_good

I also added a test for this case. I add the ref axis to one of the example rectangular FFDs under the testing files with the axis direction [0, 1, 1]. The resulting links look like this:

ref_axis_test_example

The code then tests the vectors that define the links. 

## Expected time until merged
Less than a week hopefully. This is a fairly complete feature with an associated test.

## Type of change
- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/pygeo/parameterization/DVGeo.py
+++ b/pygeo/parameterization/DVGeo.py
@@ -3213,7 +3213,8 @@ class DVGeometry(BaseDVGeometry):
             else:
 
                 if isinstance(self.axis[key]["axis"], str) and len(self.axis[key]["axis"]) == 1:
-                    # the axis can be a string of length one, then we do a ray
+                    # The axis can be a string of length one.
+                    # If so, we follow the ray projection approach.
                     if self.axis[key]["axis"].lower() == "x":
                         axis = np.array([1, 0, 0], "d")
                     elif self.axis[key]["axis"].lower() == "y":


sseraj (commented):
I tried this out on my case because my wing has some dihedral. The method works as described. I left some comments on the implementation and documentation. Please also fix the black formatting.

One other thing is that the PR description says the normal for the wing FFD is in the y-direction. Shouldn't it be the z-direction?

sseraj (dismissed):

marcomangano:
Sorry for my late review. I am still going through the code, but at a first "deeper" look I think that the current `axis` option in `addRefAxis` is too confusing. The docstring is pretty clear but I feel that we should not have a kwarg that gives you two different behaviors depending on the input being a `str` or an `array`.
What if we split `axis` into `rayaxis` and `normalaxis`, where the first accepts the string and the second the numpy array? We can either throw an error if both options are passed, or decide to have the second overwriting the first one.

marcomangano (dismissed):
I understand the issue with multiple options. I still think this is not ideal but I am fine with it, and the docstring (+ the checks) is clear enough.
I only have an additional minor comment on a docstring.

Nice work, imho this is more intuitive than the default ray projection method.

PR #179, opened 2023-01-30
Efficient interior-only embedding

sseraj:
## Purpose

I was running into long startup times when using child FFDs. The problem is that the embedding process for child FFDs can involve many points outside of the child FFD volume(s), and projecting these exterior points is expensive. In addition, projecting exterior points is more expensive with pySpline v1.5.2 than v1.5.1 because of the change in convergence criteria made in mdolab/pyspline#47.

To reduce the cost, I added a convex hull check for child FFDs (`interiorOnly=True`). One of the properties of B-splines is that they are contained within the convex hull of its control points. This means that if a point is outside the convex hull of the child FFD control points, we can skip the projection step because the point cannot be inside the FFD volume(s).

To test this, I timed the projection of a wing triangulated surface to a parent wing FFD and two child flap FFDs. The times in seconds are given in the table below:

|                 |                       | Wing | LE flap | TE flap |
|-----------------|-----------------------|------|---------|---------|
| pySpline v1.5.2 | Convex hull check     | 8.5  | 1.6     | 0.7     |
|                 | Bounding box check    | 8.5  | 19.4    | 1.7     |
|                 | No check              | 8.5  | 125.3   | 49.1    |
| pySpline v1.5.1 | Convex hull check     | 9.0  | 2.1     | 0.8     |
|                 | Bounding box check    | 8.9  | 5.3     | 0.9     |
|                 | No check              | 9.1  | 24.6    | 15.4    |

The projection times for the child FFDs are much faster with the convex hull check. The cost is similar now between pySpline versions because most exterior points are never passed to the projection routine.

I also initially tried a bounding box check, which you can see in the first commit on this branch. The bounding box is larger than the convex hull, so the projection is more expensive if the bounding box is not a close match for the FFD volume(s).

## Expected time until merged

1 week

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing

One-level FFD behavior is unchanged. The current child FFD tests pass, including the box constraint tests, which test the edge case where the FFD volume and convex hull are coincident with points lying on surface of the FFD volume. I also tested this on more realistic wing and full configuration cases.

## Checklist

### DIFF
--- a/pygeo/pyBlock.py
+++ b/pygeo/pyBlock.py
@@ -6,6 +6,7 @@ import copy
 import numpy as np
 from scipy import sparse
 from scipy.sparse import linalg
+from scipy.spatial import Delaunay
 from pyspline import Volume
 from pyspline.utils import openTecplot, writeTecplot3D, closeTecplot
 from .geo_utils import readNValues, blendKnotVectors
@@ -854,21 +855,22 @@ class pyBlock:
         v0 = 0.0
         w0 = 0.0
 
+        # If we are only interested in interior points, we skip projecting exterior points to save time.
+        # We identify exterior points by checking if they are outside the convex hull of the control points.
+        # A point can be inside the convex hull but still outside the FFD volume(s).
+        # In this case, the point is identified as an exterior point by the projection, which is more costly.
         if interiorOnly:
-            # Get the corners of the bounding box for this FFD
-            xMin, xMax = self.getBounds()
+            # Compute the Delaunay triangulation of all control points
+            triangulation = Delaunay(self.coef)
 
-            # Expand the bounds slightly in case a point lies on the boundary of the box
-            xMin -= embTol
-            xMax += embTol
+            # find_simplex returns -1 for points that are outside the Delaunay triangulation
+            # This is equivalent to points being outside the convex hull
+            isExterior = triangulation.find_simplex(x0) == -1
 
         for i in range(N):
 
-            # If we are only interested in interior points, we check if this point is outside the bounding box.
-            # If it is outside the bounding box, we skip projecting this point to save time.
-            # A point can be inside the bounding box but still outside the FFD volumes.
-            # In this case, the point is identified as an exterior point by the projection, which is more costly.
-            if interiorOnly and (any(x0[i] < xMin) or any(x0[i] > xMax)):
+            # Do not project this point if it is outside the convex hull and we are only interested in interior points
+            if interiorOnly and isExterior[i]:
                 continue
 
             for j in range(self.nVol):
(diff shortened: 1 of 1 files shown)


pygeo/pyBlock.py
```diff
@@ -858,7 +855,31 @@ def projectPoints(self, x0, checkErrors, embTol, eps, nIter):
         v0 = 0.0
         w0 = 0.0
 
+        # If we are only interested in interior points, we skip projecting exterior points to save time.
+        # We identify exterior points by checking if they are outside the convex hull of the control points.
+        # A point can be inside the convex hull but still outside the FFD volume(s).
+        # In this case, the point is identified as an exterior point by the projection, which is more costly.
```
marcomangano, line 861: Could this be rephrased to something along the lines of "Only in this case we rely on the more costly projection approach to identify the exterior points."? I got confused for a sec with the current phrasing, but probably it's just a me problem

pygeo/pyBlock.py
```diff
@@ -858,7 +855,31 @@ def projectPoints(self, x0, checkErrors, embTol, eps, nIter):
...
+
+            # ConvexHull.equations describes the planes that define the faces of the convex hull
+            # Extract the normal vector and offset for each plane
+            hullNormals = hull.equations[:, :-1]
+            hullOffsets = hull.equations[:, -1]
+
+            # The normals point outside the convex hull, so a point is inside the convex hull if the distance in the
+            # normal direction from the point to every plane defining the convex hull is negative.
+            # This is computed in a vectorized manner below.
+            distanceToPlanes = np.dot(x0, hullNormals.T) + hullOffsets
```
marcomangano, line 874: I am having a hard time getting the signs to make sense here, is the offset sign flipped?

marcomangano (commented):
Excellent, just a couple of minor questions/comments

pygeo/pyBlock.py
sseraj: Rephrased

pygeo/pyBlock.py
sseraj: For each plane, the offset is the distance (in the normal direction) from the origin to plane. The dot product by itself gives us the distance from a point to a plane that intersects the origin and has the same normal as the actual plane. Adding the offset gives us the distance to the actual plane.

sseraj: After reading the Qhull documentation a bit more and printing out the offsets for a simple case, I realized you are correct. The offset is negative in the normal direction. I added a note about this in the comments.

marcomangano (approved):

marcomangano (approved):

hajdik:
> Do we want to handle these logical checks in this PR, or merge this, and work on the logical checks separately? 

I'm fine either way. We could merge this now if there is anyone who needs this speedup on child FFDs right now, or keep all the changes together if we want one cohesive PR on embedding improvements.

Issue #176, opened 2023-01-24
DVGeometryCST import fails silently due to dependency not described in the docs


bbrelje:
## Description
If a user does not have prefoil installed, importing DVGeometryCST at pygeo init will fail silently and produce a "ImportError: cannot import name 'DVGeometryCST' from 'pygeo'" error message when the module is actually used. This hides the underlying cause, which is that prefoil is not installed. The prefoil dependency is not described, as far as I can tell, anywhere in the docs or installation instructions.

### Steps to reproduce issue
Install pygeo 1.12.2 on ubuntu-like system
Try to use DVGeometryCST (e.g. from cmplxfoil)

### Current behavior
Error message hides reason for import error

### Expected behavior
Error message ought to describe the missing dependency
Docs ought to describe prefoil depedency

### Code versions

- Operating System: Ubuntu 20
- Python: 3.8.5
- OpenMPI: n/a
- CGNS: n/a
- PETSc: n/a
- Compiler: n/a
- This repository: 1.12.2

PR #180, opened 2023-01-31
Fixing import errors, adding isort, cleaned up regression tests,  and added unit tests for new import procedure.

lamkina:
## Purpose
This PR adds the following features:
1. Improved import handling: The imports in `DVGeometryESP`, `DVGeometryVSP`, `DVGeometryMulti`, and `DVGeometryCST` all have optional dependencies required to function properly.  Previously, pyGeo suppressed the import errors leaving users confused as to why their code wasn't working with pyGeo.  In this PR, the import suppression was replaced in favor of logic control in the import statements that will throw an error if the user tries to initialize any of the classes listed above.  Using this new procedure, the correct errors will be thrown with descriptions when the user tries to use one of the classes, without suppression.
2. Added iSort: All of the import statements have been cleaned using the iSort package.  To properly format the imports with iSort, make a link from the `.isort.cfg` file [here](https://github.com/mdolab/.github/blob/main/.isort.cfg) to the root of the the pyGeo repo.  I added iSort the the azure pipeline yaml file as well.
3. Unit Test Cleaning:  As part of the changes to imports, I also fixed the inconsistencies with the regression tests and standardized how we handle imports in the tests.
4. Additional Unit Tests: I added unit tests that test the new import handling logic in the four DVGeometry classes listed above.  These tests do not require the optional dependencies because they use mock modules instead of the actual modules.

This PR will close issue #176 

## Expected time until merged
This is a quality of life PR, so not urgent.  I would estimate a few weeks or less to merge.

## Type of change

- [x] Code style update (formatting, renaming)
- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Testing

To test the new changes, run the new unit tests with either pyTest or testflo.

## Checklist


tests/unit_tests/test_import_gaurds.py
```diff
@@ -0,0 +1,98 @@
+# Standard Python modules
```
marcomangano, line 1: Is the name of this file supposed to be `test_import_guards`? Same for the class name

tests/unit_tests/test_import_gaurds.py
```diff
@@ -0,0 +1,98 @@
...
+            test_func(self, *args, **kwargs)
+
+    return do_test
+
+
+class TestImportGaurds(unittest.TestCase):
+    N_PROCS = 1
+
+    @ignore_warnings
+    def test__DVGeometryVSP_import_openvsp(self):
```
sseraj, line 21: Should be a single underscore after `test`

tests/unit_tests/test_import_gaurds.py
```diff
@@ -0,0 +1,98 @@
...
+                DVGeometryVSP("wing.vsp3")
+
+            self.assertEqual(
+                str(context.exception),
+                "The OpenVSP Python API is required in order to use DVGeometryVSP."
+                + "Ensure OpenVSP is installed properly and can be found on your path.",
+            )
+
+    @ignore_warnings
+    def test__DVGeometryVSP_openvsp_out_of_date(self):
```
sseraj, line 38: Should be a single underscore after `test`

tests/unit_tests/test_import_gaurds.py
```diff
@@ -0,0 +1,98 @@
...
+            )
+
+    @ignore_warnings
+    def test_DVGeometryMulti_import_pysurf(self):
+        with patch.dict(sys.modules, {"pysurf": None}):
+            # First party modules
+            from pygeo import DVGeometryMulti
+
+            with self.assertRaises(ImportError) as context:
+                # First party modules
```
sseraj, line 64: These should be removed when there are no imports

tests/reg_tests/test_DVConstraints.py
```diff
@@ -97,29 +104,31 @@ def generic_test_base(DVGeo, DVCon, handler, checkDerivs=True, fdstep=1e-4):
...
+        # One deforming component FFD and a stationary component FFD
+        # The components do not intersect
+        "name": "multi",
+        "child": False,
+        "multi": True,
+    },
+]
+
+# Skip multi component test if DVGeometryMulti cannot be imported (i.e. pySurf is not installed)
+@unittest.skipUnless(pysurfInstalled, "Multi-component tests require pySurf")
```
marcomangano, line 130: I believe this would skip all the tests if pysurf is not installed? The previous if statement included `self.multi` as well. Now it's working because it's a public repo so `pySurfInstalled` is always true.

I believe the logic should be `skipUnless(pysurfInstalled or not self.multi)`?

tests/unit_tests/test_import_gaurds.py
```diff
@@ -0,0 +1,98 @@
...
+            warnings.simplefilter("ignore")
+            test_func(self, *args, **kwargs)
+
+    return do_test
+
+
+class TestImportGaurds(unittest.TestCase):
+    N_PROCS = 1
+
+    @ignore_warnings
```
sseraj, line 20: Why is this necessary?

tests/reg_tests/test_DVConstraints.py
```diff
@@ -1,24 +1,31 @@
...
 
-    missing_geograd = False
+    geogradInstalled = True
 except ImportError:
-    missing_geograd = True
+    geogradInstalled = False
 
 try:
+    # First party modules
     from pygeo import DVGeometryMulti
```
sseraj, line 24: Importing the class no longer raises `ImportError`, so this check is not doing anything and the test skipping is not working. We need to check for pySurf instead like you did for others optional modules like GeoGrad above.

tests/reg_tests/test_DVGeometryMulti.py
```diff
@@ -1,23 +1,29 @@
...
 from baseclasses.utils import Error
+from mpi4py import MPI
+import numpy as np
+
+# First party modules
 from pygeo import DVGeometry
 
 try:
+    # First party modules
     from pygeo import DVGeometryMulti
```
sseraj, line 16: Need to check for pySurf as I mentioned in the DVCon tests

tests/unit_tests/test_import_gaurds.py
```diff
@@ -0,0 +1,98 @@
...
+    def test__DVGeometryVSP_import_openvsp(self):
+        with patch.dict(sys.modules, {"openvsp": None}):
```
marcomangano, line 22: This is neat

tests/unit_tests/test_import_gaurds.py
```diff
@@ -0,0 +1,98 @@
+# Standard Python modules
+import sys
+import unittest
+from unittest.mock import patch
+import warnings
+
+
+def ignore_warnings(test_func):
```
marcomangano, line 8: Nice

sseraj (changes requested):
Nice work, but it needs some changes for the test skipping to work with DVGeoMulti. I also left some more minor comments, mostly on formatting.

marcomangano (changes requested):
I think the logic for skipping tests on DVcon is not correct, but everything else looks great

tests/unit_tests/test_import_gaurds.py
marcomangano: I think this is because we are forcing the error and do not want testflo to quit - we silently ignore all the exceptions and just check them against the expected exception message

tests/reg_tests/test_DVConstraints.py
marcomangano: That's a good catch, I missed it during my review :grin:

tests/unit_tests/test_import_gaurds.py
lamkina: Are you referring to the `N_PROCS` or the `@ignore_warnings`?  The `N_PROCS` is because testflo was failing for these tests on some procs in parallel becuase of the way I temporarily modify the system packages to force the exceptions.  The `@ignore_warnings` is because numpy is imported every time we import something for these tests and it has internal warnings about efficiency that kept being raised, so I silenced them

tests/reg_tests/test_DVConstraints.py
lamkina: Good catch, I also missed this.

tests/unit_tests/test_import_gaurds.py
lamkina: This took me a long time to figure out.

tests/unit_tests/test_import_gaurds.py
lamkina: Yeah you can see my comment above, but this is because numpy is annoying for these tests.

tests/unit_tests/test_import_gaurds.py
sseraj: I was referring to the `@ignore_warnings`. I see the numpy import warning. The tests still pass without this, but I think it's fine to silence to warnings.

tests/unit_tests/test_import_gaurds.py
lamkina: Ahh yes spelling mistakes 🤦

marcomangano (approved):

sseraj (approved):

PR #186, opened 2023-02-24
Changed projectNode to work with overset surface triangulations

bmlowe4:
## Purpose
A few lines of code were added to account for the direction of the projection in projectNode (projection.py) when nsol>2. The use of overset surface meshes can lead to more than 2 resulting projections. This addesses the issue of which projected point to choose.

## Expected time until merged
A few weeks

## Type of change
- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
Create a volume and/or thickness constraint with a surface triangulation composed of overlapping meshes. 

## Checklist


pygeo/geo_utils/projection.py
```diff
@@ -218,12 +218,14 @@ def projectNode(pt, upVec, p0, v1, v2):
         # This just returns the two points with the minimum absolute
```
sseraj, line 218: My understanding of your changes is that `projectNode` now returns the closest point in the down direction and the closest point in the up direction instead of just the two closest points. If this is correct, could you update the code comments to clarify this?

pygeo/geo_utils/projection.py
```diff
@@ -218,12 +218,14 @@ def projectNode(pt, upVec, p0, v1, v2):
...
-        return newPoints[minIndex[0]], newPoints[minIndex[1]], fail
-
+        
+        # Account for the direction of the projection
+        pmin = np.dot(newPoints[:nSol] - pt, upVec)
+
+        minIndexDn = np.ma.masked_array(pmin,pmin>=0).argmin()
+        minIndexUp = np.ma.masked_array(pmin,pmin<=0).argmin()
+        
+        return newPoints[minIndexDn], newPoints[minIndexUp], fail
```
sseraj, line 228: For consistency with `nSol == 2`, I would swap the order of the returned points, and merge the check for `fail == -1` in `_generateIntersections` to be the same as `fail == 0`:
https://github.com/mdolab/pygeo/blob/b7f1bfbd01a5638dbd827013906e63498074e65b/pygeo/constraints/DVCon.py#L3309-L3315

sseraj (commented):
Thanks for opening this PR. I think this is a better way of handling more than 2 projections than what we currently have. Would you be willing to add a test for this? I can add any input files you need to our tarball and also help with writing the test.

bmlowe4:
Hey, sorry, this got put on the back burner behind some other projects.

@anilyil the CFD code we are using does not use a zipper mesh, hence our resulting surface triangulation has overlapping meshes. It is simpler to fix the node projection in DVConstraint than it is to remove the overlap in the surface triangulation. This makes the node projection algorithm more general unless you expect the two surfaces onto which you want to project to be both below or above the original plane. I don't think this will be the case based on the requirements that leList and teList be inside the wing/geometry. 

@sseraj I will work on a test for this. Apologies for the delay

sseraj:
Any updates on this @bmlowe4?

sseraj:
Hi @bmlowe4, I wanted to check on this PR again. Any updates?

sseraj:
I think we should close this PR. The incoming branch now has multiple disjoint developments.

bmlowe4:
Closed. Will continue development independently.

PR #187, opened 2023-02-27
Sparse Jacobians for DVGeometryMulti

sseraj:
## Purpose

I changed the total Jacobian computation in DVGeometryMulti to use sparse matrices instead of dense matrices. I followed one of SciPy's recommended approaches by using `lil_matrix` for construction and then converting to `csr_matrix` for faster arithmetic. This approach required minimal changes to the code.

The memory usage was not dramatically different for the cases I tested because the Jacobian is usually not very sparse. Still, the memory used by the sparse Jacobian was about 60% of the memory used by the dense Jacobian for the largest case I tested (around 140k surface points and 270 geometric DVs).

## Expected time until merged

1 week

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing

The current DVGeometryMulti test passes, which shows that the derivatives are still correct.

## Checklist

### DIFF
--- a/pygeo/parameterization/DVGeoMulti.py
+++ b/pygeo/parameterization/DVGeoMulti.py
@@ -5,6 +5,7 @@ from collections import OrderedDict
 from baseclasses.utils import Error
 from mpi4py import MPI
 import numpy as np
+from scipy import sparse
 
 try:
     # External modules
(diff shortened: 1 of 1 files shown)


marcomangano (approved):
Nice

hajdik (approved):

PR #189, opened 2023-03-06
Speed up DVJacobian computation

marcomangano:
## Purpose

Opening this PR on behalf of @anilyil.

pyGeo currently recomputes the Jacobian of the control points w.r.t. the user-defined DVs in [computeDVJacobian](https://github.com/mdolab/pygeo/blob/41ffb0ba482fa1d2ccd6336d9fde62502ad8de1d/pygeo/parameterization/DVGeo.py#L2415) every time we call [computeTotalJacobian](https://github.com/mdolab/pygeo/blob/41ffb0ba482fa1d2ccd6336d9fde62502ad8de1d/pygeo/parameterization/DVGeo.py#L2470). 
If the DVs are not updated (e.g. we are calculating derivatives on different pointSets on the same design) this step is unnecessary (and expensive) as the Jacobian remains the same.
Anil added a flag `self.JTempUpdated` to monitor if the Jacobian is up to date with the current design. If that is the case, we skip the CS-based calculation of the DVJacobian and immediately return the current matrix.
The modifications do not change the code behavior but will improve code performance, especially for more complex geometry setups.

## Expected time until merged

2-3 days

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/pygeo/parameterization/DVGeo.py
+++ b/pygeo/parameterization/DVGeo.py
@@ -172,6 +172,7 @@ class DVGeometry(BaseDVGeometry):
         # Jacobians:
         self.JT = {}
         self.nPts = {}
+        self.JTempUpdated = False
 
         # dictionary to save any coordinate transformations we are given
         self.coordXfer = {}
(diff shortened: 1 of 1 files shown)


sseraj (commented):
The implementation looks good. My only comment is about the variable name. `J_temp` is fine for a temporary local variable, but I would prefer calling it something like `J_DV` if it is going to be a class variable.

marcomangano:
> The implementation looks good. My only comment is about the variable name. `J_temp` is fine for a temporary local variable, but I would prefer calling it something like `J_DV` if it is going to be a class variable.

Agreed! Pushing soon

sseraj (approved):

PR #191, opened 2023-03-09
Fixing Constraints for new OM Parallelism Convention

bernardopacini:
## Purpose
The new OM serial derivative parallelism convention made how we handled constraints invalid as we had different derivative values on different procs for serial inputs. The solution for this is to make constraints serial and consistent on each proc. I changed all the constraints in the MPhys wrapper as well as the `compute()` and `compute_jacvec_product()` functions to reflect these changes.

## Expected time until merged
1 week.

## Type of change

- [X] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (non-backwards-compatible fix or feature)
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no API changes)
- [ ] Documentation update
- [X] Maintenance update
- [ ] Other (please describe)

## Testing
I tested this using the DAFoam MPhys aero-only and aerostructural tests. The derivatives are correct (even for TACS structural functionals) and OpenMDAO does not complain about derivative inconsistencies.

## Checklist

- [ ] I have run `flake8` and `black` to make sure the Python code adheres to PEP-8 and is consistently formatted
- [ ] I have formatted the Fortran code with `fprettify` or C/C++ code with `clang-format` as applicable
- [ ] I have run unit and regression tests which pass locally with my changes
- [ ] I have added new tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation

PR #192, opened 2023-03-13
Add triangulated surface constraint to MPhys wrapper

hajdik:
## Purpose
This adds `TriangulatedSurfaceConstraint` to `mphys_dvgeo` so packaging constraints (GeoGrad) can be used in MPhys and cleans up a few things related to that constraint in DVCon. The GeoGrad fail flag, which skips CFD if the intersection between the two objects exceeds the tolerance, has also been exposed to the MPhys wrapper using OpenMDAO's`AnalysisError`. The parallelism fix in #191 was also done for this constraint.

## Expected time until merged
1 week

## Type of change
- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
I have a runscript modified from an old MACH runscript that runs as expected with MPhys in place of MACH. If someone wants to use it in a docker container I can pass that and the input files along.

## Checklist

### DIFF
--- a/pygeo/constraints/DVCon.py
+++ b/pygeo/constraints/DVCon.py
@@ -1359,6 +1359,7 @@ class DVConstraints:
 
     def addTriangulatedSurfaceConstraint(
         self,
+        comm,
         surface_1_name=None,
         DVGeo_1_name="default",
         surface_2_name="default",
@@ -1464,6 +1465,7 @@ class DVConstraints:
 
         # Finally add constraint object
         self.constraints[typeName][conName] = TriangulatedSurfaceConstraint(
+            comm,
             conName,
             surface_1,
             surface_1_name,
--- a/pygeo/mphys/mphys_dvgeo.py
+++ b/pygeo/mphys/mphys_dvgeo.py
@@ -273,6 +273,37 @@ class OM_DVGEOCOMP(om.ExplicitComponent):
         else:
             self.add_output(name, distributed=True, shape=0)
 
+    def nom_addTriangulatedSurfaceConstraint(   
+        self,
+            name,
+            surface_1_name=None,
+            DVGeo_1_name="default",
+            surface_2_name="default",
+            DVGeo_2_name="default",
+            rho=50.0,
+            heuristic_dist=None,
+            max_perim=3.0,
+        ):
+            self.DVCon.addTriangulatedSurfaceConstraint(
+                comm=self.comm,
+                surface_1_name=surface_1_name,
+                DVGeo_1_name=DVGeo_1_name,
+                surface_2_name=surface_2_name,
+                DVGeo_2_name=DVGeo_2_name,
+                rho=rho,
+                heuristic_dist=heuristic_dist,
+                max_perim=max_perim,
+                name=name,
+            )
+
+            comm = self.comm
+            if comm.rank == 0:
+                self.add_output(f"{name}_KS", distributed=True, val=0, shape=1)
+                self.add_output(f"{name}_perim", distributed=True, val=0, shape=1)
+            else:
+                self.add_output(f"{name}_KS", distributed=True, shape=0)
+                self.add_output(f"{name}_perim", distributed=True, shape=0)
+
     def nom_addRefAxis(self, childIdx=None, **kwargs):
         # references axes are only needed in FFD-based DVGeo objects
         if self.geo_type != "ffd":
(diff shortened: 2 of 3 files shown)


eytanadler:
LGTM

sseraj (approved):

pygeo/mphys/mphys_dvgeo.py
```diff
@@ -218,16 +225,16 @@ def nom_addThicknessConstraints2D(self, name, leList, teList, nSpan=10, nChord=1
         else:
             self.add_output(name, distributed=True, shape=(0,))
 
-    def nom_addThicknessConstraints1D(self, name, ptList, nCon, axis):
-        self.DVCon.addThicknessConstraints1D(ptList, nCon, axis, name=name)
+    def nom_addThicknessConstraints1D(self, name, ptList, nCon, axis, scaled=True):
```
bernardopacini, line 228: Why is `scaled` added here? It is fine to add it, just curious about the motivation for it.

pygeo/mphys/mphys_dvgeo.py
```diff
@@ -273,6 +280,32 @@ def nom_addLinearConstraintsShape(self, name, indSetA, indSetB, factorA, factorB
         else:
             self.add_output(name, distributed=True, shape=0)
 
+    def nom_addTriangulatedSurfaceConstraint(
+        self,
+        name,
+        surface_1_name=None,
+        DVGeo_1_name="default",
+        surface_2_name="default",
+        DVGeo_2_name="default",
```
bernardopacini, line 289: Will there be an issue with these two both called "default"? I know the user will typically give names to overwrite these, but just want to make sure there won't be unexpected issues with them having the same name if they're left like this.

bernardopacini (approved):
Overall looks good to me! Just a few minor questions.

Noting that PR #191 affects this one (or vice versa, depending on which is merged first). We can go ahead and merge this and I can update that one.

pygeo/mphys/mphys_dvgeo.py
ArshSaja: I think the current formulation sets the scaled thickness constraint only. This will allow having absolute thickness constraint.

bernardopacini: Right, just wondering if there is a unique case that needs this. But, I'm totally fine with adding it regardless.

pygeo/mphys/mphys_dvgeo.py
ArshSaja: I think the test function both have surface names "default". It didn't throw any error I suppose?

pygeo/constraints/DVCon.py
```diff
@@ -1464,6 +1465,7 @@ def addTriangulatedSurfaceConstraint(
```
ArshSaja, line 1465: I think `_getSurfaceVertices` function call may cause some issues if the strings are the same because it calls the surface based on the name provided. Can the default names be different in the `mphys_dvgeo` ?

pygeo/mphys/mphys_dvgeo.py
hajdik: It was added because my test case uses the option in a thickness constraint. It's an option in DVCon but no one had needed it so it wasn't added to the MPhys layer yet.

pygeo/mphys/mphys_dvgeo.py
hajdik: This is how it is currently set up in DVCon. I can change it to a different convention but in that case I would like to keep it consistent between the two. 
Because the constraint only has to have one surface actually defined by a DVGeo in the case of an internal static object, I can either change the default for one DVGeo to None or have no default name for either DVGeo. I'm slightly against changing the default name to a string other than "default" because that's the base name for a DVGeo so you'd have to always name your DVGeo which isn't something we usually think about doing and coming up with something general for both is tricky.

pygeo/constraints/DVCon.py
hajdik: The surfaces don't have the same name, one has `surface_1_name=None` and the other has `surface_2_name="default"`.

ArshSaja: right, I mean in the mphys wrapper you define the `surface_1_name` and `surface_2_name` to be `default`. If the user doesn't change these names, will it be an issue? in your test function these surface names are different, but have you tested both surfaces' names to be `default`?

ArshSaja: Agreed, looks good to me!

ArshSaja (approved):

pygeo/mphys/mphys_dvgeo.py
ArshSaja: @bernardopacini let me know if you are okay with the feedback.

bernardopacini: I'm fine with leaving this as is since it matches the current behavior.

pygeo/mphys/mphys_dvgeo.py
bernardopacini: Sounds good! The default behavior is consistent anyways so it backwards compatibly.

bernardopacini:
I am good with this, though I cam not a pyGeo maintainer so I cannot merge it. Whoever is (@mdolab/pygeo_maintainers), please go ahead and merge this in.

PR #195, opened 2023-04-26
Reduce tolerance on tests to support Mac ARM

eirikurj:
## Purpose
This PR reduces the tolerance on few tests to get tests on the ARM based Mac Mini to pass (see https://github.com/mdolab/docker/pull/170#issuecomment-1511253227 for more details). 

## Expected time until merged
No rush, but needed for https://github.com/mdolab/docker/pull/170

## Type of change

- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Testing
Run tests. Users with M1 or M2 CPUs should preferably test this.

## Checklist

### DIFF
--- a/tests/reg_tests/test_DVGeometryMulti.py
+++ b/tests/reg_tests/test_DVGeometryMulti.py
@@ -193,7 +193,7 @@ class TestDVGeoMulti(unittest.TestCase):
         # Regression test the updated points for the real DVGeo
         refFile = os.path.join(baseDir, "ref/test_DVGeometryMulti.ref")
         with BaseRegTest(refFile, train=train) as handler:
-            handler.par_add_val("ptsUpdated", ptsUpdated, tol=1e-14)
+            handler.par_add_val("ptsUpdated", ptsUpdated, tol=1e-6)
 
         # Now we will test the derivatives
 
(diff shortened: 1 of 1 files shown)


tests/reg_tests/test_DVGeometryMulti.py
```diff
@@ -255,8 +255,8 @@ def twist(val, geo, nRefAxPts=nRefAxPts):
 
         # Check that the analytic derivatives are consistent with FD and CS
         for x in dvDict_real:
-            np.testing.assert_allclose(funcSens[x].T, funcSensFD[x], rtol=1e-4, atol=1e-10)
-            np.testing.assert_allclose(funcSens[x].T, funcSensCS[x], rtol=1e-4, atol=1e-10)
+            np.testing.assert_allclose(funcSens[x].T, funcSensFD[x], rtol=1e-4, atol=1e-7)
+            np.testing.assert_allclose(funcSens[x].T, funcSensCS[x], rtol=1e-4, atol=1e-7)
```
sseraj, line 259: I have to double check this, but from what I remember, this derivative test will be essentially useless with such a loose `atol`

sseraj (changes requested):

tests/reg_tests/test_DVGeometryMulti.py
eirikurj: Yeah, the FD tests could be close... If that is the case, we either need platform dependent refs, or just not test on ARM for now. Let me know, but we can also discuss on https://github.com/mdolab/docker/pull/170 or offline.

sseraj: I checked this and there are entries in `funcSens` that are order 1e-8 and 1e-9. These are also the entries that are nonzero because of the intersection treatment, so loosening `atol` to 1e-7 means that we would not be testing the intersection derivatives. It's possible that the derivatives are wrong on the ARM image.

bernardopacini: I don’t have an M1/M2 unfortunately… @eytanadler @hajdik @awccopp do.

bernardopacini: @eytanadler is editing my comment to remove himself from this list…

bernardopacini: @hajdik too…

eytanadler: I tried installing the necessary packages to test this locally but ran into problems compiling installing pySurf on ARM. I don't have Docker set up to test these easily, sorry.

PR #197, opened 2023-05-01
change extra flags test check to compiler

eirikurj:
## Purpose
New Intel Ubuntu docker images do not have ESP and OpenVSP installed, resulting in the tests failing, as no tests are allowed to skip (see https://github.com/mdolab/docker/pull/193). This PR updated the check to the compiler instead of the OS.

## Expected time until merged
Should be quick once tests pass.

## Type of change

- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/.github/test_real.sh
+++ b/.github/test_real.sh
@@ -2,8 +2,8 @@
 set -e
 ./input_files/get-input-files.sh
 
-# All tests should pass on Ubuntu
-if [[ $OS == "ubuntu" ]]; then
+# All tests should pass on GCC
+if [[ $COMPILERS == "gcc" ]]; then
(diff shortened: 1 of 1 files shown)


eirikurj:
TACC image failed before because previously the `OS` variable was not set on this image, thus skipped test are allowed. ESP is not installed because of an Intel MPI check when building the docker images. To be consistent, I added `I_MPI_ROOT` check also for Intel MPI. Should be good now, but I any case we should get ESP/OpenVSP integrated on Intel.

marcomangano (approved):

sseraj (approved):

PR #199, opened 2023-05-10
DVGeoMulti test fix

anilyil:
## Purpose

The inverse distance warping math in DVGeoMulti uses a small number "eps" to prevent some denominators from being zero or getting a wrong sign. Previously, this clipping was not done in a good way, and the eps value was selected to be too small.  See the diff in this PR for the relevant code in DVGeoMulti.

In this PR, I changed the clipping to more accurately represent the computational goal, increased eps from 1e-50 to 1e-20, and retrained the tests. 

This is done to address the issue discussed in this PR: https://github.com/mdolab/pygeo/pull/195

This tolerance issue is coming up on this test because some test points are defined to be _exactly_ on the intersection curve. This is a worst-case scenario in terms of numerical accuracy and this is what the eps-based clipping was trying to prevent. Real cases with intersection curves that are not straight are very unlikely to get CFD mesh points that are exactly on the curve. As a result, this change should not affect any real life  application. 

Increasing eps is the main mechanism that fixes this issue for the worst-case test point. However, while I was at it, I also fixed the clipping math so that its more accurate. 

Finally, with the eps value of 1e-50, the ARM and X86 architectures were getting very different answers because of the numerical instability in the algorithm. Slight changes in the updated point coordinates were causing large changes in derivatives as expected. The increased eps value fixes these and the updated points are now consistent to around 10 digits instead of 4-5 with the original eps value.

## Expected time until merged

1-2 days

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist


marcomangano (approved):
Wild catch! Let's merge this asap (@eirikurj  we need once again you privileges due to the RTD failures) so we can finally close #195

eirikurj:
Overall looks good. @anilyil are these tests trained on `amd64` or `arm64`? Just want to check, since we do want to favor `amd64` architecture. Just let me know and I can push updated values if you want.

pygeo/parameterization/DVGeoMulti.py
```diff
@@ -1508,8 +1508,8 @@ def update(self, ptSetName, delta):
             # Compute denominators for the integral evaluations
             # Add an epsilon so that these terms never become zero
```
sseraj, line 1509: Looks good, but maybe this comment should also be updated slightly now that we use min?

marcomangano (approved):

sseraj (approved):

PR #193, opened 2023-03-28
LocalSectionDVs in MPhys wrapper

hajdik:
## Purpose
The MPhys wrapper was missing local section DVs. 

## Expected time until merged
2 days

## Type of change
- [ ] Bugfix (non-breaking change which fixes an issue)
- [ x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (non-backwards-compatible fix or feature)
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no API changes)
- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
soon

## Checklist
- [ ] I have run `flake8` and `black` to make sure the Python code adheres to PEP-8 and is consistently formatted
- [ ] I have formatted the Fortran code with `fprettify` or C/C++ code with `clang-format` as applicable
- [ ] I have run unit and regression tests which pass locally with my changes
- [ ] I have added new tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation

PR #201, opened 2023-05-17
Add names to child DVGeos

anilyil:
## Purpose

This PR changes how DVGeo class tracks its children. Old implementation just appended the children to a list, which is then indexed in order. The current implementation saves the child DVGeometries in the `self.children` dictionary. The keys of the dictionary are the names of each child DVGeo. This makes the code a lot more readable since each child DVGeo is referenced by its name. 

The old functionality is maintained, and unless users were accessing the child DVGeometries in a hacky way by indexing the old list, everything should work fine. Some output names might be different, but we can also fix this behavior. The names of the added pointsets will be the same if users don't provide a name for the child DVGeo.

I will apply a similar set of changes to the mphys wrapper once #193 is merged. 

## Expected time until merged

2 weeks

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist


pygeo/parameterization/DVGeo.py
```diff
@@ -752,6 +752,11 @@ def coordXfer(coords, mode="fwd", applyDisplacement=True, **kwargs):
         # compNames is only needed for DVGeometryMulti, so remove it if passed
         kwargs.pop("compNames", None)
 
+        # check if we want a custom subset of child DVGeos
```
marcomangano, line 755: This looks like an interesting addition, could you please add a short description of the arg and intended functionality in the docstrings for `addPointSet`?

marcomangano (commented):
I gave a quick pass and it looks ok to me, see my comment below. Maybe we should also add a test for `activeChildren`?

bernardopacini (approved):
The changes here look good to me, just a question regarding compatibility:

- How does this affect other DVGeo variants such as DVGeoMulti or DVGeoAxi? If I am not mistaken, it should be fine because they inherit this base anyways, right?

hajdik:
> Should we change the approach here to rely on that name attribute?

Yes, there is a `name` attribute in the BaseDVGeo class so everything inheriting from that has access to it, I think it makes sense to use that since it's already there.

tests/reg_tests/test_DVGeometry.py
```diff
@@ -1460,6 +1460,72 @@ def test_shape_functions(self, train=False, refDeriv=False):
...
+                dIdPt = np.zeros([nPt, 2, 3])
+                dIdPt[0, 0, 0] = 1.0
+                dIdPt[1, 0, 1] = 1.0
+                dIdPt[2, 0, 2] = 1.0
+                dIdPt[3, 1, 0] = 1.0
+                dIdPt[4, 1, 1] = 1.0
+                dIdPt[5, 1, 2] = 1.0
+                dIdx = DVGeo.totalSensitivity(dIdPt, ptName)
+
+                handler.root_add_dict(f"dIdx_{ptName}", dIdx, rtol=1e-10, atol=1e-10)
```
hajdik, line 1517: Looks like the test is failing here with `KeyError: 'child1_span1'` but once that's fixed I'm good with this PR

pygeo/parameterization/DVGeo.py
```diff
@@ -908,17 +941,28 @@ def addGlobalDV(self, dvName, value, func, lower=None, upper=None, scale=1.0, co
             Use a string for a single configuration or a list for multiple
             configurations. The default value of None implies that the design
             variable applies to *ALL* configurations.
+        prependName : bool
+            Flag to determine if self.name attribute is prepended to this DV name.
```
hajdik, line 945: Can you add something in the docstring about why this option exists?

hajdik (approved):

PR #207, opened 2023-07-11
Multi-DVGeo capability in MPhys

hajdik:
## Purpose
This adds the ability to add multiple DVGeo objects to one MPhys geometry component when setting up the `OM_DVGEOCOMP` while preserving backward compatibility for the more common single DVGeo case.
DVGeometry classes were also modified to standardize `name` and `type` attributes to enable tracking of multiple DVGeos. 

## Expected time until merged
1 week

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
In addition to the MACH pyGeo tests passing, I have some test scripts for single-DVGeo and multi-DVGeo MPhys in place of actual MPhys tests.

## Checklist


pygeo/mphys/mphys_dvgeo.py
```diff
@@ -84,23 +130,29 @@ def compute(self, inputs, outputs):
...
+    def nom_addChild(self, ffd_file, DVGeoName=None):
+        # if we have multiple DVGeos use the one specified by name
+        if self.multDVGeo:
+            DVGeo = self.DVGeos[DVGeoName]
+        else:
+            DVGeo = self.DVGeo
+
         # can only add a child to a FFD DVGeo
-        if self.geo_type != "ffd":
+        if DVGeo.geoType != "ffd":
```
sseraj, line 141: One suggestion: Instead of adding `geoType` to each parameterization class, use `isinstance` to check what class `DVGeo` is. For example, this line would be `if not isinstance(DVGeo, DVGeometry):`. This is less error prone than using strings.

pygeo/mphys/mphys_dvgeo.py
```diff
@@ -84,23 +130,29 @@ def compute(self, inputs, outputs):
...
             raise RuntimeError(
-                f"Only FFD-based DVGeo objects can have children added to them, not type:{self.geo_type}"
+                f"Only FFD-based DVGeo objects can have children added to them, not type:{DVGeo.geoType}"
```
sseraj, line 143: Similarly, `DVGeo.geoType` here can be replaced with `type(DVGeo)` or `type(DVGeo).__name__`, which will be just the class name.

pygeo/mphys/mphys_dvgeo.py
hajdik: Good point, I updated these to all directly check `type`

hajdik:
> Maybe one simplifying pattern could be to always initialize the dvgeos as if you have multiple, but if dvgeo info is not passed, you can just populate that with the single dvgeo's information. 

I implemented a version of this and tested it out, let me know what you think @anilyil

ArshSaja:
My derivatives are fine. The only thing is that if I need to get some values from the DVGeo in my transcript, I would just use `self.geo.DVGeo.getValues()` earlier. But now I need to call it with the default DVGeo name: `self.geo.DVGeos["defaultDVGeo"].getValues()`.  I think you should explain this in the MPhys wrapper.

hajdik:
> The only thing is that if I need to get some values from the DVGeo in my runscript, I would just use `self.geo.DVGeo.getValues()` earlier. But now I need to call it with the default DVGeo name: `self.geo.DVGeos["defaultDVGeo"].getValues()`. 

I think if you want to call a function that isn't in the MPhys wrapper like `getValues()`, the best way would be to use  `nom_getDVGeo()` to get the DVGeo object and then call any functions directly on that.

lamkina (approved):
I tested this with my aeroprop case and everything looks good.  LGTM.

PR #209, opened 2023-07-13
Add ability for writeSurfaceTecplot to get surface from DVGeo

A-CGray:
## Purpose

The `writeSurfaceSTL` in `DVConstraints` has the ability to write out a surface that is not stored in a `DVConstraints` object but in the associated DVGeo. This PR just extends that functionality to the `writeSurfaceTecplot` function.

## Expected time until merged

1 day

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/pygeo/constraints/DVCon.py
+++ b/pygeo/constraints/DVCon.py
@@ -359,7 +359,7 @@ class DVConstraints:
             self.linearCon[key].writeTecplot(f)
         f.close()
 
-    def writeSurfaceTecplot(self, fileName, surfaceName="default"):
+    def writeSurfaceTecplot(self, fileName, surfaceName="default", fromDVGeo=None):
         """
         Write the triangulated surface mesh used in the constraint object
         to a tecplot file for visualization.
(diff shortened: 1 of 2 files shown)


sseraj (commented):
The name `_getSurfacePoints` is too similar to `_getSurfaceVertices`. Can you modify `_getSurfaceVertices` to add the `fromDVGeo` argument instead of adding a new function?

sseraj (approved):

A-CGray:
> Changes look good. One minor thing: can you add what the default behavior is when `fromDVGeo` is set to `None`? Does it just write the original coordinates? To someone looking at the current state of the code w/o any of its history, it would be difficult to understand the different behavior.

Sure, please check that my explanation makes sense to you. As far as I can see, if `fromDVGeo=None` then the points are taken from the DVCon's `self.surfaces` dict, which only ever contains the original surface coordinates. Whereas, if you supply a DVGeo name, it will retreive the current deformed coordinates from DVGeo

PR #198, opened 2023-05-01
Updating DVGeoVSP to work with OpenVSP 3.33.0

timryanb:
## Purpose
A backwards-incompatible change has been introduced in OpenVSP starting with version 3.33.0 (see release notes [here](https://github.com/OpenVSP/OpenVSP/releases/tag/OpenVSP_3.33.0)). This change affects the scaling of one of the parameter coordinates used in OpenVSP's volume parameterization. The fix to this is pretty straight-forward on the pyGeo side: include a conditional scale factor on the affected coordinate depending on which OpenVSP version is installed.

## Expected time until merged
1 week

## Type of change

- [ ] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (non-backwards-compatible fix or feature)
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no API changes)
- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Checklist

- [x] I have run `flake8` and `black` to make sure the Python code adheres to PEP-8 and is consistently formatted
- [ ] I have formatted the Fortran code with `fprettify` or C/C++ code with `clang-format` as applicable
- [x] I have run unit and regression tests which pass locally with my changes
- [ ] I have added new tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation

PR #210, opened 2023-07-14
Moved OpenVSP version check to fix API docs

sseraj:
## Purpose

#198 broke the API docs for DVGeometryVSP, the modules that follow it in `__init__.py` (DVGeometryESP, DVGeometryMulti), and the MPhys wrapper. The problem is that the `openvsp.GetVSPVersion()` call occurs outside the class, which does not work when we have mock imports for the doc build. To fix this, I moved the version check into the `__init__` function.

## Expected time until merged

1 week

## Type of change

- [ ] Documentation update
- [x] Maintenance update
- [ ] Other (please describe)

## Testing

Check the API docs for the latest docs build and compare to the PR build. All the API docs should be working on the PR build.

## Checklist

### DIFF
--- a/pygeo/parameterization/DVGeoVSP.py
+++ b/pygeo/parameterization/DVGeoVSP.py
@@ -30,22 +30,6 @@ except ImportError:
         vspInstalled = False
 
 
-vspOutOfDate = False
-if vspInstalled:
-    vspVersionStr = openvsp.GetVSPVersion()
-    words = vspVersionStr.split()
-    vspVersion = words[-1]
-    # VSP is installed, but version too old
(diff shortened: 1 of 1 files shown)


hajdik (approved):
LGTM but @ArshSaja or @lamkina should make sure VSP still works with one of their cases

lamkina (changes requested):
@sseraj  there is a unit test for the out of date features, but can you expand it to check the different version conditions?  I think the machinery is there to do it fairly easily.

lamkina:
Okay, I guess it's fine.  We're throwing a descriptive error message about the version so the user will be made aware of the problem.

lamkina (approved):

PR #229, opened 2023-11-29
Add proximity constraints

anilyil:
## Purpose

This PR adds "proximity constraints", which are regular thickness constraints but they can be added between different surfaces, and the points tracked on each surface can have a custom set of argument when they are added to the DVGeo object. This enables us to have a bit more control over thickness constraints that are added in areas where multiple surfaces might be kicking around.

This feature was developed on a branch fron "children_dict", so the commit history is long, but the changes should be minor. I added tests for the expected behavior.

## Expected time until merged

1-2 weeks

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist


pygeo/constraints/DVCon.py
```diff
@@ -764,6 +764,191 @@ def addThicknessConstraints1D(
...
+            determine the intersections with each component surface.
+
+        vecList : list or array of size (N x 3)
+            Directions that originate from each point in ptList
+            that define the search vector directions for component
+            surface intersections. The vector direction should
+            point to component A, and the opposite direction should
+            point to component B
+
+        surfA : str
```
hajdik, line 802: Are surfA and surfB the names of the surfaces held in DVCon?

pygeo/constraints/DVCon.py
```diff
@@ -764,6 +764,191 @@ def addThicknessConstraints1D(
...
+            Name of the surface mesh that defines component A
+
+        surfB : str
+            Name of the surface mesh that defines component B
+
+        pointSetKwargsA : dict
```
hajdik, line 808: Does this constraint need pointSetKwargs to work or can the proximity be between the entire surface?

hajdik (approved):
Looks great, just a few comments/questions

pygeo/constraints/DVCon.py
```diff
@@ -764,6 +764,191 @@ def addThicknessConstraints1D(
...
+        lower=1.0,
+        upper=3.0,
+        scaled=True,
+        scale=1.0,
+        name=None,
+        addToPyOpt=True,
+        DVGeoName="default",
+        compNames=None,
+    ):
+        r"""
```
lamkina, line 784: Any objections to adding a disclaimer to the docstring that says something like:

`
Surfaces must move in the same direction as the constraint orientation to prevent skewness. i.e, if the toothpicks are oriented in the y-direction, the surfaces must only move in the y-direction.
`
</code>

I can see a scenario where two objects both shear and move normal w.r.t each other and these constraints are skewed.  I know this a symptom of the underlying toothpick constraints, but I think it's much easier to achieve this effect with two separate pointsets moving around.

lamkina (changes requested):
Just one comment on documenting the limitations of the constraint.

lamkina (approved):

hajdik (approved):

PR #230, opened 2023-12-02
Adding DVGeometryMulti to MPhys

bernardopacini:
## Purpose

Adding DVGeometryMulti to MPhys. I added pass-through functions and modified the `compute` and `compute_jacvec_product` calls as needed.

## Expected time until merged
1 week.

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
I tried this on my optimization and it appears to work (local analysis and derivatives on HPC). I would appreciate others verifying this on their cases.

## Checklist


bernardopacini:
I didn't add the `addIntersection()` method to the MPhys wrapper. Let me know if you think it should be here or if it should be left as a direct reference to the underlying class.

bernardopacini:
Additional note: when adding a global DV the input was being added twice to OpenMDAO. This bug is fixed here too.

bernardopacini:
It turns out PR #201 broke the naming convention for child FFDs by prepending the child name by default. This meant that MPhys would not connect the OM variable to the DVGeo variable because the names would not match. I fixed that in this PR by setting `prependName=False` for all FFD-based design variables.

sseraj (commented):
Looks pretty good. I just have a few minor comments

sseraj (dismissed):

pygeo/mphys/mphys_dvgeo.py
```diff
@@ -125,7 +128,21 @@ def compute(self, inputs, outputs):
         # next time the jacvec product routine is called
         self.update_jac = True
 
-    def nom_addChild(self, ffd_file, DVGeoName=None, childName=None):
+    def nom_addComponent(self, comp, ffd_file, triMesh, DVGeoName=None):
```
hajdik, line 131: Can you add a docstring for this function since you added it??

pygeo/mphys/mphys_dvgeo.py
```diff
@@ -205,14 +225,22 @@ def nom_getDVGeo(self, childName=None, DVGeoName=None):
...
-        else:
+        # return a component of a DVGeoMulti
+        elif comp is not None:
+            return DVGeo.DVGeoDict[comp]
+
+        # return a child of a DVGeo
+        elif childName is not None:
             return DVGeo.children[childName]
 
+        # return the top level DVGeo
```
hajdik, line 240: I agree with Anil on using this to cover finding the right DVGeo even with the child/component logic instead of repeating the same checks. I think I just didn't realize it could eliminate the checks for child that I put in when I redid the DV adding functions.

hajdik (commented):
LGTM, just a few comments

bernardopacini:
There is some weird behavior in the derivatives: the first check totals is fine, but a second check totals after a design perturbation makes the second function derivative wrong. This only happens when there is an intersection. I can't exclude this being an issue with DAFoam (or even IDwarp), but I do not have the bandwidth to look more into this now. Converting to a draft.

pygeo/mphys/mphys_dvgeo.py
bernardopacini: With the recent changes to DVGeo, someone took care of this update already. All the checks should be streamlined now.

pygeo/mphys/mphys_dvgeo.py
bernardopacini: Added.

bernardopacini:
This is good to go. The issues with derivatives are coming from other code components, not DVGeo, and they are not unique to this PR. We should merge this in so it does not go stale again.

bernardopacini:
> Looks good overall. Is there a corresponding addIntersection call? If not, how are you handling it?

Instead of having a pass-through call in MPhys, I just call the DVGeo method directly. For example:

```        
DVGeo = self.geometry.nom_getDVGeo()
self.geometry.nom_addComponent("hub", "FFD/hubFFD.xyz", triMesh="hubTri.cgns")
self.geometry.nom_addComponent("blade", "FFD/bladeFFD.xyz", triMesh="bladeTri.cgns")
DVGeo.addIntersection(
    "hub",
    "blade",
    dStarA=1.0,
    dStarB=0.05,
    featureCurves=featureCurves,
    project=True,
    includeCurves=True,
    curveEpsDict=curveEpsDict,
    anisotropy=[1.0, 1.0, 0.75],
    slidingCurves=slidingCurves,
    trackSurfaces=trackSurfaces,
)
```

Is this fine with you?

pygeo/mphys/mphys_dvgeo.py
```diff
@@ -125,7 +128,39 @@ def compute(self, inputs, outputs):
         # next time the jacvec product routine is called
         self.update_jac = True
 
-    def nom_addChild(self, ffd_file, DVGeoName=None, childName=None):
+    def nom_addComponent(self, comp, ffd_file, triMesh, DVGeoName=None):
```
hajdik, line 131: Do we need this pass through function to modify anything on the OM side or could everything be handled with DVGeo calls in a runscript?

bernardopacini: I thought about this too and debated what to do. The user would have to import DVGeo and make a DVGeo object to then add it to the DVGeoMulti object. It would work and I can change it to that, but this call streamlines the process. I agree that if the goal is to include only things that actually modify OM in the MPhys wrapper then we should change this.

hajdik: That's fair, I think in this case it's cleaner to handle it with this function.

hajdik (approved):

PR #231, opened 2023-12-02
Adding sliding curves for DVGeoMulti

bernardopacini:
## Purpose

There are cases when using the intersection method in which a curve on a surface should be preserved to make sure that the patch does not change size or the curve is maintained to a certain shape. For example, maintaining the interface line between a propeller hub and nacelle is important to ensure a realistic boundary condition on each patch. This PR adds this sort of tracking.

These changes are actually made by @anilyil.

## Expected time until merged
2 weeks.

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing
I have run this locally and also checked the derivatives to make sure the code is correct.

## Checklist

### DIFF
--- a/pygeo/parameterization/DVGeoMulti.py
+++ b/pygeo/parameterization/DVGeoMulti.py
@@ -184,6 +184,7 @@ class DVGeometryMulti:
         project=False,
         marchDir=1,
         includeCurves=False,
+        slidingCurves=[],
         intDir=None,
         curveEpsDict=None,
         trackSurfaces=None,
@@ -232,6 +233,10 @@ class DVGeometryMulti:
 
         includeCurves : bool, optional
             Flag to specify whether to include features curves in the inverse-distance deformation.
+
+        slidingCurves : list, optional
+            The list of curves to project to, but on which the mesh nodes are not frozen in their initial positions.
+            This allows the mesh nodes to slide along the feature curve.
 
         intDir : int, optional
             If there are multiple intersection curves, this specifies which curve to choose.
@@ -290,6 +295,7 @@ class DVGeometryMulti:
                 project,
                 marchDir,
                 includeCurves,
+                slidingCurves,
                 intDir,
                 curveEpsDict,
                 trackSurfaces,
@@ -1006,6 +1012,7 @@ class CompIntersection:
         project,
         marchDir,
         includeCurves,
+        slidingCurves,
         intDir,
         curveEpsDict,
         trackSurfaces,
@@ -1194,6 +1201,10 @@ class CompIntersection:
 
         # flag to include feature curves in ID-warping
         self.incCurves = includeCurves
+
+        # list of curves that allow nodes to slide on them. we only use these for the projection step,
+        # but they are not included in the first line based IDWarp
+        self.slidingCurves = slidingCurves
 
         # direction to pick if we have multiple intersection curves
         self.intDir = intDir
@@ -1433,7 +1444,7 @@ class CompIntersection:
                     elemIDs[:] = elemIDs + 1
                     # (we need to do this separetely because Fortran will actively change elemIDs contents.
                     self.curveSearchAPI.mindistancecurve(
-                        ptsToCurves.T, self.seam0.T, self.seamConn.T + 1, xyzProj.T, tanProj.T, dist2, elemIDs
+                        ptsToCurves.T, self.seam0.T, self.seamConnFull.T + 1, xyzProj.T, tanProj.T, dist2, elemIDs
                     )
 
                     # Adjust indices back to Python standards
@@ -1543,7 +1554,7 @@ class CompIntersection:
         # we use the initial seam coordinates here
         coor = self.seam0
         # bar connectivity for the remeshed elements
-        conn = self.seamConn
+        conn = self.seamConnWarp
         # deltas for each point (nNode, 3) in size
         if self.seam.shape == self.seam0.shape:
             dr = self.seam - self.seam0
@@ -1650,7 +1661,7 @@ class CompIntersection:
         # we use the initial seam coordinates here
         coor = self.seam0
         # bar connectivity for the remeshed elements
-        conn = self.seamConn
+        conn = self.seamConnWarp
 
         # Get the two end points for the line elements
         r0 = coor[conn[:, 0]]
@@ -1787,7 +1798,7 @@ class CompIntersection:
             # conn of the current curve
             seamBeg = self.seamBeg[curveName]
             seamEnd = self.seamEnd[curveName]
-            curveConn = self.seamConn[seamBeg:seamEnd]
+            curveConn = self.seamConnFull[seamBeg:seamEnd]
 
             # Project these to the combined curves using pySurf
             # Get number of points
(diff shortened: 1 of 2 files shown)


bernardopacini:
I added a test for the sliding curve feature as discussed offline with @hajdik and @sseraj. It checks to ensure that a few points remain on the lines on which they were defined, but allowed to vary along them (@anilyil let me know if it is okay with you too).

tests/reg_tests/test_DVGeometryMulti.py
```diff
@@ -315,6 +315,157 @@ def twist(val, geo, nRefAxPts=nRefAxPts):
...
+
+        # Define the communicator
+        comm = MPI.COMM_WORLD
+
+        # Set up real component DVGeo objects
+        DVGeoBox1 = DVGeometry(ffdFiles[0])
+        DVGeoBox2 = DVGeometry(ffdFiles[1])
+
+        # Set up real DVGeometryMulti object
+        DVGeo = DVGeometryMulti(comm=comm, debug=True)
```
sseraj, line 332: `debug` should be False so that running the test doesn't generate `.plt` files

tests/reg_tests/test_DVGeometryMulti.py
```diff
@@ -315,6 +315,157 @@ def twist(val, geo, nRefAxPts=nRefAxPts):
...
+                for i in range(1, nRefAxPts):
+                    geo.rot_z["box"].coef[i] = val[i - 1]
+
+            DVGeoDict[comp].addGlobalDV(dvName=f"{comp}_twist", value=[0] * nTwist, func=twist)
+
+        # Add the point set
+        pts_dtype = localPts
+        DVGeo.addPointSet(pts_dtype, ptSetName, comm=comm, applyIC=True)
+
+        # Apply twist to the two intersecting boxes
```
sseraj, line 441: Only one box is twisted

pygeo/parameterization/DVGeoMulti.py
```diff
@@ -2892,7 +2904,10 @@ def _getIntersectionSeam(self, comm, firstCall=False):
 
                 # append this new curve to the featureCurve data
                 remeshedCurves = np.vstack((remeshedCurves, newCoor))
-                remeshedCurveConn = np.vstack((remeshedCurveConn, newBarsConn))
+                remeshedCurveConnFull = np.vstack((remeshedCurveConnFull, newBarsConn))
+
+                if curveName not in self.slidingCurves:
```
sseraj, line 2909: By excluding sliding curves here, they are not used as seeds for the curved-based deformation. This means that points on these curves get warped like any other point. We also still want the 'full' connectivity because that is used for projections. Is this correct? A comment or two about this would be helpful

pygeo/parameterization/DVGeoMulti.py
```diff
@@ -1195,6 +1202,10 @@ def __init__(
         # flag to include feature curves in ID-warping
         self.incCurves = includeCurves
 
+        # list of curves that allow nodes to slide on them. we only use these for the projection step,
+        # but they are not included in the first line based IDWarp
```
sseraj, line 1206: What does "first" here mean? Also, would it be more accurate to say these curves are not included as seeds in the curve-based deformation?

pygeo/parameterization/DVGeoMulti.py
```diff
@@ -184,6 +184,7 @@ def addIntersection(
         project=False,
         marchDir=1,
         includeCurves=False,
+        slidingCurves=[],
```
sseraj, line 187: To avoid mutable defaults, this should be initialized to None and then set to empty in `__init__` like `featureCurves`

sseraj (commented):
This is a great feature. Thanks for adding the test as well. I have a few minor comments

pygeo/parameterization/DVGeoMulti.py
bernardopacini: I will defer to @anilyil on this.

pygeo/parameterization/DVGeoMulti.py
bernardopacini: Again, I will defer to @anilyil on this.

bernardopacini:
Thanks for the feedback @sseraj! I updated the files with your comments and defer to @anilyil on the implementation-specific comments.

tests/reg_tests/test_DVGeometryMulti.py
```diff
@@ -315,6 +315,157 @@ def twist(val, geo, nRefAxPts=nRefAxPts):
...
+        comm.Allgatherv(sendbuf, recvbuf)
+
+        # Reshape into a nPtsGlobal, 3 array
+        ptsUpdated = globalPoints.reshape((nPtsGlobal, 3))
+
+        # Test that the X and Z coordinates are unchanged and Y coordinates are changed
+        for i in range(np.size(pts, 0)):
+            self.assertAlmostEqual(pts[i, 0], ptsUpdated[i, 0])
+            self.assertNotEqual(pts[i, 1], ptsUpdated[i, 1])
+            self.assertAlmostEqual(pts[i, 2], ptsUpdated[i, 2])
```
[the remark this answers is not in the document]
sseraj: I was thinking on something like this:
```
# Test that the X and Z coordinates are unchanged and Y coordinates are changed
np.testing.assert_allclose(pts[:, 0], ptsUpdated[:, 0], rtol=, atol=)
np.testing.assert_allclose(pts[:, 2], ptsUpdated[:, 2], rtol=, atol=)

with self.assertRaises(AssertionError):
    np.testing.assert_allclose(pts[:, 1], ptsUpdated[:, 1], rtol=, atol=)
```
The tolerances should be tight for the first two and loose for the last one

bernardopacini:
> @sseraj, I addressed both of your comments. Also, thank you for spelling out pretty much all of the changes needed in the comments. Both things you said are correct. Please review my changes and mark the conversations as resolved.
> 
> @bernardopacini, this PR is good for another round of reviews, correct?

I just updated the assertions as recommended by @sseraj, now should be good to go.

tests/reg_tests/test_DVGeometryMulti.py
```diff
@@ -315,6 +315,139 @@ def twist(val, geo, nRefAxPts=nRefAxPts):
...
+
+        # Allgather the updated coordinates
+        comm.Allgatherv(sendbuf, recvbuf)
+
+        # Reshape into a nPtsGlobal, 3 array
+        ptsUpdated = globalPoints.reshape((nPtsGlobal, 3))
+
+        # Test that the X and Z coordinates are unchanged and Y coordinates are changed
+        np.testing.assert_allclose(pts[:, 0], ptsUpdated[:, 0], rtol=1e-4, atol=1e-10)
+        np.testing.assert_allclose(pts[:, 2], ptsUpdated[:, 2], rtol=1e-4, atol=1e-10)
```
sseraj, line 446: The `rtol` for these should be tighter than 1e-4, right? I would expect the coordinates to match to 8 digits at least

tests/reg_tests/test_DVGeometryMulti.py
```diff
@@ -315,6 +315,139 @@ def twist(val, geo, nRefAxPts=nRefAxPts):
...
+
+        with self.assertRaises(AssertionError):
+            np.testing.assert_allclose(pts[:, 1], ptsUpdated[:, 1], rtol=1e-4, atol=1e-10)
```
sseraj, line 449: This `rtol` is fine because the points should differ by more than 1e-4

tests/reg_tests/test_DVGeometryMulti.py
bernardopacini: I'll make it tighter. I matched the other tests for consistency but I'll change it, you are right that it should match essentially to numerical precision.

sseraj (approved):

PR #241, opened 2024-04-05
Various small changes

A-CGray:
## Purpose

- Changes the hard-coded lower bound for the triangulated surface constraint from a dummy value of `-1e10` to `None` which is the correct way to represent that there is no lower bound.
- Converts the SVD vectors computed when using composite DVs from a numpy matrix to a numpy array. Using a matrix causes the design variable values to end up as a 2D array which causes pyoptsparse to fail.
- Fixes a check in `addRefAxis` that is supposed to evaluate to false is x/y/zFraction is None, but currently also evaluates to False is the values are zero.

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/pygeo/parameterization/DVGeo.py
+++ b/pygeo/parameterization/DVGeo.py
@@ -1520,6 +1520,7 @@ class DVGeometry(BaseDVGeometry):
             self.computeTotalJacobian(ptSetName)
             J_full = self.JT[ptSetName].todense()  # this is in CSR format but we convert it to a dense matrix
             u, s, _ = np.linalg.svd(J_full, full_matrices=False)
+            u = np.array(u)
             scale = np.sqrt(s)
             # normalize the scaling
             scale = scale * (NDV / np.sum(scale))
--- a/pygeo/parameterization/DVGeo.py
+++ b/pygeo/parameterization/DVGeo.py
@@ -1518,9 +1518,8 @@ class DVGeometry(BaseDVGeometry):
             if ptSetName is None:
                 raise ValueError("If u and s need to be computed, you must specify the ptSetName")
             self.computeTotalJacobian(ptSetName)
-            J_full = self.JT[ptSetName].todense()  # this is in CSR format but we convert it to a dense matrix
+            J_full = self.JT[ptSetName].toarray()  # this is in CSR format but we convert it to a dense matrix
             u, s, _ = np.linalg.svd(J_full, full_matrices=False)
-            u = np.array(u)
             scale = np.sqrt(s)
             # normalize the scaling
             scale = scale * (NDV / np.sum(scale))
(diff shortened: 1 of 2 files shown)


pygeo/parameterization/DVGeo.py
```diff
@@ -1520,6 +1520,7 @@ def addCompositeDV(self, dvName, ptSetName=None, u=None, scale=None, prependName
             self.computeTotalJacobian(ptSetName)
             J_full = self.JT[ptSetName].todense()  # this is in CSR format but we convert it to a dense matrix
             u, s, _ = np.linalg.svd(J_full, full_matrices=False)
+            u = np.array(u)
```
sseraj, line 1523: The other two changes look good. Can you explain this one a bit more? I thought `u` was supposed  to be an `N_DV` by `N_DV` matrix

A-CGray: Yeah it is, the issue is that it's an N_DV by N_DV `np.matrix` not `np.array`. Because of this, `values` ends up being a (1,265) `np.matrix` and that breaks pyoptsparse. Calling flatten on a matrix doesn't do anything because the matrix type always wants to remain 2d.

sseraj (dismissed):

pygeo/parameterization/DVGeo.py
A-CGray: @ArshSaja , I think you're the only person other than @ewu63 who's used the composite DVs, did you ever run into this?

pygeo/parameterization/DVGeo.py
```diff
@@ -506,7 +506,7 @@ def addRefAxis(
                     "rot0axis": rot0axis,
                 }
             nAxis = len(curve.coef)
-        elif xFraction or yFraction or zFraction:
+        elif xFraction is not None or yFraction is not None or zFraction is not None:
```
eirikurj, line 509: The checks are good, but there are other if checks also later.

pygeo/parameterization/DVGeo.py
A-CGray: Oh, good catch, shows you how observant I am. I agree, will change to `toarray` and remove this conversion

pygeo/parameterization/DVGeo.py
A-CGray: Done

pygeo/parameterization/DVGeo.py
sseraj: @ArshSaja confirmed that he also had to fix this to use composite DVs, so it seems like this was broken from the start

sseraj (approved):

pygeo/parameterization/DVGeo.py
ArshSaja: I had similar issue in the past. Since this is local change (`todense()` to `toarray()` in that function, it won't affect anywhere else I think. It looks good to me

PR #253, opened 2024-09-18
Fix `evalFunctionsSens` for projected area constraint

A-CGray:
## Purpose

The `evalFunctionsSens` method of the projected area constraint currently fails if it is called when there are no geometric design variables because only the creation of the dAp0/1/2 arrays are skipped, but the rest of the function, that uses those arrays, is not. This is fixed by simply skipping everything if `nDV ==0`.

Also:
- Altered the projected area constraint to allow for arbitrary projection directions (not just x,y,z)
- Made the sensitivity calculation marginally more efficient
- Split the box projected area test into one test that checks that the projected areas in the x y and z direction match the expected values, and another that tests the derivatives of the projected area in a random direction. The reason is that the sensitivities of the box's projected areas in the x, y and z directions are inaccurate due to the presence of surfaces perfectly orthogonal to the projection direction
- Added some comments in the code explaining this issue with the derivatives
- Switched from using `np.random.seed` and `np.random.rand` to `np.random.default_rng` and `rng.random`
- Re-trained DVConstraints tests due to new random numbers

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist


marcomangano:
The changes make sense, but I wonder if it is more elegant to have something like:
```
if nDV == 0:
      pass
```

A-CGray:
> The changes make sense, but I wonder if it is more elegant to have something like:
> 
> ```
> if nDV == 0:
>       pass
> ```

I don't think it matters, this solution would be 1 line longer than what's currently there.

marcomangano (dismissed):

A-CGray:
@eytanadler so glad we figured this out so now you and @hajdik can get the mphys tests working :)

tests/reg_tests/test_DVConstraints.py
```diff
@@ -668,11 +668,14 @@ def test_surfaceArea_box(self, train=False, refDeriv=False):
             )
 
     def test_projectedArea(self, train=False, refDeriv=False):
+        np.random.seed(37)
```
eytanadler, line 671: Very minor, but NumPy suggests that it is better practice to use their generator rather than these functions to support legacy code. The suggested method creates a generator with `gen = np.random.default_rng(37)` and then generating random numbers with `gen.random()`.

tests/reg_tests/test_DVConstraints.py
```diff
@@ -719,6 +732,17 @@ def test_projectedArea_box(self, train=False, refDeriv=False):
                 atol=1e-7,
             )
 
+    def test_projectedArea_box_funcSens(self, train=False, refDeriv=False):
+        np.random.seed(37)
```
eytanadler, line 736: Great test to add, thanks. Same minor comment about the random number generator as previously.

eytanadler (approved):
Excellent fixes, thank you @A-CGray (in particular for getting to the bottom of those derivatives!

tests/reg_tests/test_DVConstraints.py
A-CGray: Yeah you're right, I was just being lazy and copying what's done elsewhere in these tests, will change it now

A-CGray: The `default_rng.random` doesn't produce the same output as the old `np.random.rand` even with the same seed, so I had to re-train all the tests

eytanadler: Well that’s frustrating….thanks for updating

eytanadler (approved):

hajdik (approved):

PR #261, opened 2025-02-13
Fix `nom_addLocalSectionDV` in MPhys wrapper

eytanadler:
## Purpose

The parameters to `addLocalSectionDV` in the MPhys wrapper are passed without keyword arguments, which means they default to the order the keyword arguments were defined. As a result, the variables were getting mixed up when `addLocalSectionDV` was called and the answer was wrong.

This also fixes adding pointsets to children, which happens already in `DVGeometry.addPointSet` and doesn't need to be repeated in `OM_DVGEO.nom_addPointSet`.

## Expected time until merged

1 day

## Type of change

- [x] Bugfix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (non-backwards-compatible fix or feature)
- [ ] Code style update (formatting, renaming)
- [ ] Refactoring (no functional changes, no API changes)
- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist

- [ ] I have run `flake8` and `black` to make sure the Python code adheres to PEP-8 and is consistently formatted
- [ ] I have formatted the Fortran code with `fprettify` or C/C++ code with `clang-format` as applicable
- [ ] I have run unit and regression tests which pass locally with my changes
- [ ] I have added new tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation

PR #263, opened 2025-03-03
Parallel test fix

A-CGray:
## Purpose

In #261 we switched all the parallel tests to run on as many procs as are available. However, some of the ESP tests are designed to run on a specific number of cores and hang indefinitely on our buildbots when they're run with a different number of cores. So this PR fixes the number of procs used for those tests, and sets the correct environment variables to allow MPI oversubscription so that the tests will still work if we have fewer cores available.

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/tests/reg_tests/test_DVGeometryESP.py
+++ b/tests/reg_tests/test_DVGeometryESP.py
@@ -391,7 +391,7 @@ class TestPyGeoESP_BasicCube(unittest.TestCase):
 @unittest.skipUnless(mpiInstalled and ocsmInstalled, "MPI and pyOCSM are required.")
 class TestPyGeoESP_BasicCube_Distributed(unittest.TestCase):
     numPhysicalCores = psutil.cpu_count(logical=False)
-    N_PROCS = int(np.clip(numPhysicalCores, 1, 16))
+    N_PROCS = 3
 
     def setUp(self):
         # Store the path where this current script lives
@@ -590,7 +590,7 @@ class TestPyGeoESP_BasicCube_Distributed(unittest.TestCase):
 @unittest.skipUnless(mpiInstalled and ocsmInstalled, "MPI and pyOCSM are required.")
 class TestPyGeoESP_BasicCube_Distributed_OneProcBlank(unittest.TestCase):
     numPhysicalCores = psutil.cpu_count(logical=False)
-    N_PROCS = int(np.clip(numPhysicalCores, 1, 16))
+    N_PROCS = 4
 
     def setUp(self):
         # Store the path where this current script lives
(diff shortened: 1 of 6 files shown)


tests/reg_tests/test_MPhysGeo.py
```diff
@@ -34,6 +34,34 @@
     ocsmInstalled = False
 
 
+def assert_check_totals(totals, atol, rtol):
```
eirikurj, line 37: This change seems unrelated to this PR

tests/reg_tests/test_DVGeometryESP.py
```diff
@@ -391,7 +391,7 @@ def test_composite(self, train=False):
 @unittest.skipUnless(mpiInstalled and ocsmInstalled, "MPI and pyOCSM are required.")
 class TestPyGeoESP_BasicCube_Distributed(unittest.TestCase):
     numPhysicalCores = psutil.cpu_count(logical=False)
```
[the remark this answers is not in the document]
eirikurj: Can you also update the `parallel_4procs` text in [this line](https://github.com/mdolab/pygeo/blob/300843253a4cdd7ba860450cb47682a8f8a3ec28/tests/reg_tests/test_DVGeometryESP.py#L39) since its does not match what was updated in https://github.com/mdolab/pygeo/pull/261

tests/reg_tests/test_MPhysGeo.py
A-CGray: Right, not sure how this got in here, how can I remove this commit without undoing the changes?

A-CGray: nvm, this was actually code that should've been deleted

eirikurj (commented):
Thanks for addressing the comments. However, looking over the changes in https://github.com/mdolab/pygeo/pull/261/files again I think there are few issues and inconsistencies
* Many tests have both serial and parallel tests. For the parallel tests I think the lines like `N_PROCS_CUR = int(np.clip(numPhysicalCores, 1, 16))` should have a lower bound of 2 not 1 core, otherwise we are not testing in parallel/distributed. 
* The string `parallel_4procs` still shows up in several other tests.

eirikurj:
We should change the other tests as well
`test_DVGeometryMulti.py`:
-  still has the procs set to 1 (see https://github.com/mdolab/pygeo/blob/300843253a4cdd7ba860450cb47682a8f8a3ec28/tests/reg_tests/test_DVGeometryMulti.py#L33)
- the string is inconsistent (see https://github.com/mdolab/pygeo/blob/300843253a4cdd7ba860450cb47682a8f8a3ec28/tests/reg_tests/test_DVGeometryMulti.py#L40)

`test_DVGeometryCST.py`:
- still has the procs set to 1 (see https://github.com/mdolab/pygeo/blob/300843253a4cdd7ba860450cb47682a8f8a3ec28/tests/reg_tests/test_DVGeometryCST.py#L328)

eirikurj (approved):

PR #264, opened 2025-03-04
error handling on mphys pygeo 'type'

gawng:
## Purpose

This places an else check on the info['type'] because I kept passing in `type` keyword argument as uppercase 'FFD' when it was looking for lowercase 'ffd'. Now at least the warning message will tell the user what is breaking.

## Expected time until merged

Relatively simple so maybe a few days?

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing

I ran it locally passing in 'FFD' and it returned an error instead of proceeding.

## Checklist

### DIFF
--- a/.github/test_real.sh
+++ b/.github/test_real.sh
@@ -7,5 +7,12 @@ if [[ $COMPILERS == "gcc" ]] && [[ -z $I_MPI_ROOT ]] && [[ "$(arch)" == "x86_64"
     EXTRA_FLAGS='--disallow_skipped'
 fi
 
+# Set OpenMPI env variables only on non-Intel MPI
+if [[ -z $I_MPI_ROOT ]]; then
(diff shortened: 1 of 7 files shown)


marcomangano (commented):
This looks good, but what if we cast `info["type"]` to lower case by default, maybe [here](https://github.com/mdolab/pygeo/blob/77bd90d197863a7a973dfae628c7bcf31f8f7eb8/pygeo/mphys/mphys_dvgeo.py#L49)? Not sure if that reflect best coding practices though

marcomangano (dismissed):

A-CGray (changes requested):
Why not check the equality using `info["type"].lower()`? That way it's case insensitive.

A-CGray (approved):

marcomangano (approved):

PR #280, opened 2025-10-10
New thickness to chord constraint implementation

A-CGray:
## Purpose

The current thickness to chord constraint implementations has a few limitations:

1. You could only compute thickness to chord along a 1D line.
2. The constraint didn't actually compute an accurate chord length, it just used two points spaced slightly apart in the chordwise direction. This meant you could only use the thickness to chord constraint in the scaled sense (relative to the baseline).

This PR adds a completely new implementation of t/c constraints in DVConstraints. The new implementation basically uses two sets of thickness constraints, one to compute the thicknesses and another to compute the chord lengths.

ToCConstraints

This new implementation has the following advantages:

1. You can compute thickness to chord over a 1D line or a 2D grid
2. The constraints use leading and trailing edge points to actually compute an accurate chord length and therefore accurate (non-scaled) t/c values
3. The 2D version can compute the max t/c value for each chordwise section using KS aggregation

Technically these changes are backward incompatible because defining the 1D thickness to chord constraints now requires a set of leading and trailing edge coordinates instead of a projection direction and a chord direction. I don't think this is a big issue because these constraints weren't used by many people.

## Type of change

- [ ] Documentation update
- [ ] Maintenance update
- [ ] Other (please describe)

## Testing

I've expanded the existing thickness to chord constraint test to test the new capabilities (2D, section max)

## Checklist

- [x] I have run `ruff check` and `ruff format` to make sure the Python code adheres to PEP-8 and is consistently formatted


sabakhshi (approved):

A-CGray:
@marcomangano do you have some time to give this a look over? If not I can also ask Hannah

marcomangano:
> @marcomangano do you have some time to give this a look over? If not I can also ask Hannah

I can take a look by Mon/Tue the latest!

A-CGray:
@wkupiec reminder about this

wkupiec (approved):
lgtm

